PR #4, opened 2019-02-22
Add docs/tutorial for Bitbots Motion

timonegk:


SammyRamone:
added some small things. Looks good otherwise, can be merged in my oppinion

jgueldenstein:
I've made some small corrections and additions.

Maybe its better to have this in English for the future

PR #20, opened 2019-08-13
Vision debug

NFiedler:
This is a very small script which enables the debug output of the vision with dynamic reconfiguration. I think it is very useful because usually, the vision is running on a robot without debug output enabled. 
I did not add any ".sh" stuff because I'm thinking about adding the scripts directory to the PATH variable which would allow us to call it and other scripts easily in the terminal.


NFiedler:
Yes, i did.

ftsell:
Something I just remembered as well: Why is this not in the _scripts_ folder inside _bitbots_vision_?

SammyRamone:
@NFiedler ping

timonegk:
This script should not be necessary, bit-bots/bitbots_vision#116 should be used instead.

PR #23, opened 2019-08-15
updated architecture dia

SammyRamone:
updated the architecture diagramm. there are still issues open, layout is also not very pretty currently. Would be good if some people have a look at it to see if somethings wrong or missing


SammyRamone:
will solve https://github.com/bit-bots/bitbots_meta/issues/18

SammyRamone:
> I would like to see the UDP bridge somewhere in there.
> 
> It doesn't necessarily have to be connected to all the topics which it actually subscribes because that would just be confusing. But having it in the testing or visualization tools would still make sense.

done

ftsell:
It might be me but I don't see it. Did you amend your previous commit? Because I also don't see a new commit

SammyRamone:
> It might be me but I don't see it. Did you amend your previous commit? Because I also don't see a new commit

hm, something went wrong :/ but now it should be there :)

timonegk:
`humanoid_league_transform` subscribes to `/camera_info`; that connection is missing.

Also, the `white_balancer` should be between `pylon_camera` and `/image_raw`. The topic between `pylon_camera` and `white_balancer` is `/image_resized`.

And why are the `pressure_converter` and the `complementary_filter` in different shapes?

SammyRamone:
> `humanoid_league_transform` subscribes to `/camera_info`; that connection is missing.
> 
> Also, the `white_balancer` should be between `pylon_camera` and `/image_raw`. The topic between `pylon_camera` and `white_balancer` is `/image_resized`.
> 
fixed your requested changes.


> And why are the `pressure_converter` and the `complementary_filter` in different shapes?
Because complementary_filter is a package that we use from somewhere else, but pressure_converter is currently a script in the ros_control package (which will be refactored). If I would use a node shape, the script would fail to find the package "pressure_converter".

timonegk:
Okay, I see.
As far as I know, @JanGut is currently reorganizing the dia. We should postpone the merge if this is still the case.

JanGut:
Refactoring of the architecture wolfgang.dia for version 1 is now finished.

PR #63, opened 2019-12-12
Use python 3 in scripts

timonegk:
## Proposed changes
This sets the python version to 3 when the workspace is installed and when building on the robot.

## Necessary checks
- [ ] ~~Update package version~~
- [ ] Run linters
- [ ] Run `catkin build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/scripts/install.pl
+++ b/scripts/install.pl
@@ -71,7 +71,7 @@ sub create_catkin_workspace() {
     mkdir $location;
     chdir $location;
     mkdir $location . "/src";
-    system "bash -c \"source $ROS_WORKSPACE/setup.bash && catkin init && catkin build\"";
+    system "bash -c \"source $ROS_WORKSPACE/setup.bash && catkin init && catkin config -DPYTHON_VERSION=3 && catkin build\"";
 }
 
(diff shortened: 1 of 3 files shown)


scripts/robot_compile.py
```diff
@@ -359,7 +359,7 @@ def build(target, package='', pre_clean=False):
          "cd {workspace};"
          "source devel/setup.zsh;"
          "{cmd_clean}"
-         "catkin build --force-color {package} {quiet_option} --continue-on-failure --summary || exit 1;"
+         "catkin build --force-color {package} {quiet_option} --continue-on-failure --summary -DPYTHON_VERSION=3 || exit 1;"
```
ftsell: I think we should configure the robots workspace to always have this option just like any developers normal environment.
This would prevent accidents when someone only adds a simple change directly on the robot and calls `catkin build` without the explicit option.

timonegk: Yes, you are right. I removed the option and configured all robot workspaces accordingly.

PR #101, opened 2021-06-12
Also pull RoboCupProtocol in pull_all

timonegk:


ftsell (commented):
Why not use `git submodule foreach --recursive` ?

timonegk:
The submodule's submodule is currently not checked out to master and instead on a commit, therefore `git pull` would not work. I could change that but I don't think it's bad to have the correct commit of the submodule in the team communication, because the team communication's code directly depends on a specific version of the protocol.

ftsell (approved):

PR #104, opened 2021-10-12
Remove ros_pathes from sync includes

jaagut:
## Proposed changes
Introduced in #103.
Remove `bitbots_ros_patches` as it is not needed anymore.


Flova:
Aren't the ros patches still needed for the cpp code through?

jaagut:
There are only python ros_patches. The problem with `Rate` existed only in the cpp implementation and was fixable without patches as it only required listening to some return value.

timonegk (approved):

PR #112, opened 2022-04-13
Make pull-all acually pulls everything now

jaagut:
`make pull-all` now also pulls the bitbots_meta


timonegk:
This is already done I will the Makefile.

PR #114, opened 2022-04-14
Switch back to upstream tf2_geometry_msgs

Flova:
## Proposed changes
- Remove PyKDL build from source
- Remove tf2_geometry_msgs


jaagut:
We are removing this because we switched to releases or what is the reason behind this?

Flova:
The tf2 upstream package should be available again, so we can use that. I still need to look if it is only released for jammy.

Flova:
My docker is jammy, so I am not sure if the robot receives the update.

jaagut:
Seems like `tf2_geometry_msgs` is not available in 20.04
@val-ba thanks for investigating!

jaagut (approved):

PR #209, opened 2023-10-25
Fix balser driver and make setup easier

Flova:
# Summary
- The basler repo was updated and we need to upgrade the basler driver version otherwise the build fails
- Add the blaze support package as it is now required for the ros driver. The package is only for basler time of flight cameras, but sadly it is not optional.
- Don't pull the drivers from a private repo
- Add direct download, which is possible because the download URLs are quite clear and don't seem to change. (we are not the only ones on GitHub who seem to do that...)
- Add some checks that provide helpful information if e.g. the URL is outdated and returns a 404.


scripts/make_basler.sh
```diff
@@ -1,8 +1,69 @@
...
+# Go to the download button and copy the link address.
+PYLON_DOWNLOAD_URL="https://www2.baslerweb.com/media/downloads/software/pylon_software/pylon_7_4_0_14900_linux_x86_64_debs.tar.gz"
+PYLON_VERSION="7.4.0"
+
+# Similar to the pylon driver we also need to download the blaze supplementary package.
+BLAZE_DOWNLOAD_URL="https://www2.baslerweb.com/media/downloads/software/tof_software/pylon-supplementary-package-for-blaze-1.5.0.def07388_amd64.deb"
+BLAZE_VERSION="1.5.0"
+
+# Check let the user confirm that they read the license agreement on the basler website and are fine with it.
+echo "You need to confirm that you read the license agreements for pylon $PYLON_VERSION and the blaze supplementary package $BLAZE_VERSION on the basler download page (https://www.baslerweb.com/en/downloads/software-downloads/) and are fine with it."
```
jaagut, line 15:
```suggestion
# Check let the user confirm that they read the license agreement on the basler website and agree with it.
echo "You need to confirm that you read the license agreements for pylon $PYLON_VERSION and the blaze supplementary package $BLAZE_VERSION on the basler download page (https://www.baslerweb.com/en/downloads/software-downloads/) and agree with it."
```

scripts/make_basler.sh
```diff
@@ -1,8 +1,69 @@
...
+
+# Check -ci flag for automatic confirmation in the ci
+if [[ $1 == "-ci" ]]; then
+    echo "Automatic confirmation detected. Continuing..."
```
jaagut, line 19:
```suggestion
    echo "Running in a CI environment, continuing..."
```

scripts/make_basler.sh
```diff
@@ -1,8 +1,69 @@
...
+    echo "Pylon download url does not exist. Please check the url and update the 'PYLON_DOWNLOAD_URL' variable in the 'make_basler.sh' script. The website might have changed."
+    exit 1
+fi
+
+if ! curl --output /dev/null --silent --head --fail "$BLAZE_DOWNLOAD_URL"; then
+    echo "Blaze download url does not exist. Please check the url and update the 'BLAZE_DOWNLOAD_URL' variable in the 'make_basler.sh' script. The website might have changed."
+    exit 1
+fi
+
+# Check if the correct pylon driver PYLON_VERSION is installed (apt)
```
jaagut, line 47: Please check this before testing for internet connection. This way you can exit successfully even without connectivity.

jaagut (approved):

PR #210, opened 2023-10-27
Bump ros setup action version

Flova:
# Summary
Currently, there is a warning due to an old node version being used for the ros setup part. This hopefully fixes that.


Flova:
Let's wait until the ci passes at least once.

jaagut (approved):

PR #219, opened 2023-11-29
Use Makefile instead of make_* scripts

timonegk:
# Summary
Currently, scripts that call new scripts are used for dependencies in our `make ...` commands, but this is already supported by `Makefile`. This PR removes these scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,18 @@
 basler:
 	scripts/make_basler.sh $(ARGS)
 
-install: pull-init
-	scripts/make_update.sh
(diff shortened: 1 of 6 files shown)


jaagut (commented):
And the master command tries to switch to the branches defined in submodule or master?

timonegk:
Exactly, it is the same command that is also part of pull_init. It switches to the branch specified as the default branch in the gitmodules file or to master if no branch is given.

jaagut (approved):

PR #222, opened 2024-01-18
Create Monorepo

Flova:
# Summary
This PR creates a monorepo for all bitbots related software. ROS sports and 3rd party libs are still in other repos, but this is fine and can be managed with vcs tool. All git history is preserved! CI as well as repo handling should be much easier that way. What to you think?

In case we want to go this route we need to:
- Adjust the tooling
    - Fix docs ci
    - Fix pull-all etc scripts
    - Fix sync includes
    - Fix meta ci
- Archive old repos
- Add note to old repos
- Add vcs tool config
- Migrate issues / open PRs


Flova:
Do I get something for the biggest ever PR in bit-bots history? :P

timonegk:
We originally decided to have multiple repositories because it makes dealing with git and conflicts easier and facilitates independent work on different parts of the software. I think these reasons are still valid. This is a huge change and I am not sure if it really improves things.

Flova:
While I see some of your points, I would disagree with others. 

Merge conflicts should not be an issue if you don't change the same files. And if you do, it is the same if you have multiple repos. But if you have multiple ones you have multiple gits you need to keep track of. Also, tracking which branch is needed in which repo (for cross repo changes) is quite the manual process and prone to errors. Similarly, you don't need to open ~10 pull requests for the same change, if you e.g. rename a message. Similarly, the tooling can be simplified a lot and is less confusing for new people. Using a more standardized approach also makes it easier for people to find help (googling how to do something with git is easier than seeing why a custom script fails because you are on the wrong branch in some submodule).
Having a bunch of changed submodules as the default state in the meta isn't ideal ether as it might be confusing, or it hides accidental changes. Easily reaching a clean master/main can be quite helpful. 

Also, if you get started it is quite hard to get an overview, where you need to open a PR and navigate GitHub. Having this many repos increases this. Also, seeing the changes of other people on the same project might help when you try to get to know the different components. And as much as we try to isolate everything, in the end integration is important, and you often need to change things in many repos while integrating or tuning. Making this harder is not something we want.

I understand that having separate repos is beneficial if you want to share code with other people. But the code in this repo is relatively tightly coupled. Isolated parts with much more stable APIs are released into ROS sports in their own repositories or as binaries. If somebody still decides to use, e.g. our walking. They will probably fork and adapt it either way to fit their needs. 

Working with a singular repo for both SAR Cam and the master project (8 people with quite isolated subsystems) was quite refreshing tbh.. Also, most other teams as well as most large projects (with also include many isolated subsystems) use this approach.

timonegk:
That makes sense. I think the original problem with merge conflicts / merge commits was from before we used pull requests. Then, every `git pull` created a new merge commit locally. But that should no longer be an issue.
You are probably right about the advantages. I think we could give it try. Maybe decouple some parts of the code that can be used by others, for example the DSD.

jaagut:
Merges since creation of this PR:

- [ ] https://github.com/bit-bots/bitbots_vision/pull/359

Flova:
Changes after creation (formatting, CI adjustments, vcs tool, docs):

https://github.com/bit-bots/bitbots_meta/compare/c0c716e...monorepo

Flova:
Now we also run the tests

jaagut (approved):

jaagut:
We now want to switch to a monorepo workflow. As this PR contains practically every code line (over 90,000!) it is not feasible to review everything. However, this represents the old sub-repo state plus some additional tests and tooling changes. Those seem to be fine.
Thus, we are now ready to progress.

PR #339, opened 2024-02-07
Add fast deployment

Flova:
# Summary
- Reuses the locally compiled binaries on the robot
- based on #223

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot
- [ ] This PR is on our `Software` project board


scripts/deploy/deploy_robots.py
```diff
@@ -99,6 +100,12 @@ def _parse_arguments(self) -> argparse.Namespace:
         )
 
         # Optional arguments
+        parser.add_argument(
+            "-f",
+            "--fast",
+            action="store_true",
+            help="Don't build/compile on the target machines, instead synchronize the complete install directory from the local machine.",
```
texhnolyze, line 107: I would write build/install, as we are syncing both as far as I've seen.

scripts/deploy/deploy_robots.py
```diff
@@ -99,6 +100,12 @@ def _parse_arguments(self) -> argparse.Namespace:
         )
 
         # Optional arguments
+        parser.add_argument(
```
texhnolyze, line 103: I personally would suggest renaming or even skipping the `-f` shorthand, because in a lot of tools `-f` stands for `force` and breaks with general CLI tooling convention.

Also, as long as we are not using the `-f` flag for multiple combined speedup options, I would suggest defining a `dest` with a clear variable name, as with the other arguments.

E.g. `sync_build_alternative` or just `skip_build` or something of the sort.

scripts/deploy/deploy_robots.py
```diff
@@ -140,6 +147,12 @@ def _parse_arguments(self) -> argparse.Namespace:
             args.build = args.only_build
             args.launch = args.only_launch
 
+        if args.fast:
```
texhnolyze, line 150: I think it's generally fine, but now you could theroretically pass both `--build` and `--fast` and get a potentially confusing outcome for the user, as changed files on the robot are also cleaned and no build is happening.

scripts/deploy/tasks/sync.py
```diff
@@ -10,28 +10,31 @@
...
         :param pre_clean: Whether to clean the source directory before syncing
+        :param fast: Whether to sync the current local colcon workspace including build artifacts
         """
         super().__init__()
-        self._local_workspace = local_workspace
+        self._local_src_bitbots_main = local_src_bitbots_main
         self._remote_workspace = remote_workspace
         self._package = package
         self._pre_clean = pre_clean
+        self._fast = fast
```
texhnolyze, line 35: I think `fast` is especially wrong in a `Sync` class, seeing as the non fast sync includes less files and should actually be faster.

scripts/deploy/tasks/sync.py
```diff
@@ -84,8 +87,34 @@ def _run(self, connections: Group) -> GroupResult:
...
+                os.path.dirname(os.path.dirname(self._local_src_bitbots_main))
+                + "/",  # NOTE: The trailing slash is important for rsync
+                [
+                    "'--include=- log'",
+                    "'--include=+ build'",
+                    "'--include=+ build/**'",
+                    "'--include=+ install'",
+                    "'--include=+ install/**'",
+                    "'--include=+ src'",
+                    "'--include=+ src/**'",
```
texhnolyze, line 102: Could including all `src/**` subfolders lead to later build/execution issues, when things are copied that are not part of our sync includes?

texhnolyze (changes requested):
Hey sorry, I missed that the PR is still a draft.
Feel free to disregard stuff, that is not applicable anymore.

jaagut:
We don't want to finish this, as the feature "fast deployement" my introduce errors and we can only use this in a few situations. Current methods work sufficiently. We don't need to introduce this additional dirty complexity.

Issue #177, opened 2023-03-08
Predeclare and validate parameters by YAML


texhnolyze:
As in ros2 parameters have to be predeclared we want to configure them by YAML configuration while allowing validation and bounds, by utilizing: https://github.com/PickNikRobotics/generate_parameter_library

While the library generates C++ classes for python we need to implement our own approach, by potentially generating a class/object at runtime.

The follwoing steps must be done:
- introduces the `generate_parameter_library` lib
- implement code/object generation for python
- setup YAML configuration of parameters for all modules in `bitbots_meta`
- adjust code of modules to predeclare parameters utilizing this approach
- rm backwards compatible flags `allow_undelcared_parameters` and `declare_parameters_from_overrides` from all modules

jaagut:
From our meeting notes: Use only one YAML file if only one parameter set exists

Originally posted: https://github.com/bit-bots/bitbots_meta/issues/204#issue-1880466698

jaagut:
This tool has already been introduced in the `bitbots_template_repository`. However the following parameter-files need to be replaced:

84 files: 
- [ ] `bitbots_behavior/bitbots_body_behavior/config/animations.yaml`
- [ ] `bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml`
- [ ] `bitbots_lowlevel/bitbots_buttons/config/buttons.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/analyzers.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/pressure_amy.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/pressure_converter.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/pressure_melody.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/pressure_nobot.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/pressure_rory.yaml`
- [ ] `bitbots_lowlevel/bitbots_ros_control/config/wolfgang.yaml`
- [ ] `bitbots_misc/bitbots_basler_camera/config/camera_settings.yaml`
- [ ] `bitbots_misc/bitbots_basler_camera/config/default.yaml`
- [ ] `bitbots_misc/bitbots_ceiling_cam/config/camera_settings_ceiling_cam.yaml`
[...]

PR #342, opened 2024-02-08
New param style

ayin21:


Flova:
I still think using the Gen parameter lib in the behavior would be cool.

jaagut:
We still track this topic in #177. However, this PR is really outdated and we can reopen and update this at any time or do it in a new PR.

PR #349, opened 2024-02-19
Remove robot controllers

Flova:
# Summary

The library is not used in the code at all. So I think it is not needed anymore. Also, it throws some warnings, is not maintained and throws warnings.

Accidentally based on #348.


jgueldenstein (approved):
lgtm

PR #379, opened 2024-04-17
feat(localization): add german_open default field

texhnolyze:
# Summary
Add field configuration for German Open.


jaagut (changes requested):
Please call it garman_open2024 instead to differentiate it in the future

PR #387, opened 2024-04-17
Remove odometry measurement time offset

Flova:
# Summary
- Remove measurement timing correction using odom


jaagut (approved):
In talking to you, we were not 100% certain, this has no side-effects. Perhaps, we just want to disable this instead of removing? But I guess, this is a clean change and we can still revert...

Flova:
> In talking to you, we were not 100% certain, this has no side-effects. Perhaps, we just want to disable this instead of removing? But I guess, this is a clean change and we can still revert...

I would also think that we can just revert it if we need to

PR #405, opened 2024-04-19
Consistent `game_settings.yaml` over deployment syncs

texhnolyze:
# Summary
The `game_settings.yaml` were added to the excludes for syncing, and we now use a merge of a dict of defaults with the current `game_settings.yaml` from the last configuration.
This has the effect, that if the settings were already configured, they can stay the same and the user only has to accept.
Additionally at the end of the configuration process, the user is now shown the current configuration, that will be written and has to accept or can choose to redo the process with a `[Y/n]` prompt.

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot

### DIFF
--- a/bitbots_misc/bitbots_utils/bitbots_utils/game_settings.py
+++ b/bitbots_misc/bitbots_utils/bitbots_utils/game_settings.py
@@ -15,6 +15,12 @@ SETTING_PATH = os.path.join(
     "config",
     "game_settings.yaml",
 )
+DEFAULT_SETTING_PATH = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))),
+    "bitbots_utils",
+    "config",
+    "default_game_settings.yaml",
+)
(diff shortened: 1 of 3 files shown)


bitbots_misc/bitbots_utils/config/game_settings.yaml
```diff
@@ -1,8 +1,8 @@
 parameter_blackboard:
   ros__parameters:
     bot_id: 1
+    monitoring_host_ip: 172.20.0.10
```
jaagut, line 4: Where does this come from?

jaagut (approved):

bitbots_misc/bitbots_utils/config/game_settings.yaml
texhnolyze: I thought it might make sense to just set a default monitoring IP in our internal network. Since so far we have been using `*.*.*.10` in games as well, I just chose this one.

PR #409, opened 2024-04-19
Estimate yaw for localization reinit after fall

Flova:
# Summary
Store IMU orientation when we fall. Check the IMU yaw after the fall and estimate how much the robot has rotated during the fall. This leads to a better re-initialization especially when we fall on ambiguous locations like the center point.

## Proposed changes
- Store imu yaw 
- Reinit based on change in imu yaw

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_navigation/bitbots_localization_handler/bitbots_localization_handler/localization_dsd/actions/initialize.py
+++ b/bitbots_navigation/bitbots_localization_handler/bitbots_localization_handler/localization_dsd/actions/initialize.py
@@ -1,3 +1,5 @@
+import math
+
 import tf2_ros as tf2
 from bitbots_localization.srv import ResetFilter
 from rclpy.duration import Duration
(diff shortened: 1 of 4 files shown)


JanNiklasFeld (approved):

bitbots_navigation/bitbots_localization_handler/bitbots_localization_handler/localization_dsd/actions/initialize.py
```diff
@@ -68,6 +70,34 @@ def perform(self, reevaluate=False):
         return self.pop()
 
 
+class InitPoseAfterFall(AbstractInitialize):
+    def perform(self, reevaluate=False):
+        self.do_not_reevaluate()
```
texhnolyze, line 75: I think non reevaluated actions should be marked somehow in the naming, that it is clear in `*.dsd` files, that they stop reevaluation till completion.

Does not need to be done now, but then we should create an issue.

PR #415, opened 2024-04-20
Limit unrealistically large motion updates

Flova:
# Summary
If e.g. during a fall a large rotation is output by the odometry or if a component like the odometry is reseted the localization should ignore the motion update as it otherwise leads to "exploding particles" because the unrealistically large updates leads to a lot of added noise.

## Proposed changes
- Add parameters with a max velocity for the motion updates

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_navigation/bitbots_localization/config/config.yaml
+++ b/bitbots_navigation/bitbots_localization/config/config.yaml
@@ -5,6 +5,8 @@ bitbots_localization:
       percentage_best_particles: 50
       min_motion_linear: 0.0
       min_motion_angular: 0.0
+      max_motion_linear: 0.5
+      max_motion_angular: 3.14
       filter_only_with_motion: false
     ros:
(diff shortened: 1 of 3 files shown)


Flova:
TODO test on robot [ ]

bitbots_navigation/bitbots_localization/config/config.yaml
```diff
@@ -5,6 +5,8 @@ bitbots_localization:
       percentage_best_particles: 50
       min_motion_linear: 0.0
       min_motion_angular: 0.0
+      max_motion_linear: 0.5
+      max_motion_angular: 3.14
```
texhnolyze, line 9: A comment might be nice if radian/degrees are used, if the parameter is ever changed.

texhnolyze (approved):

PR #419, opened 2024-04-20
Fix/dynup back better

JanNiklasFeld:
# Summary

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
@JanNiklasFeld What/why did you change the spline in the engine?

Flova:
> @JanNiklasFeld What/why did you change the spline in the engine?

I think it moves the head at a different point, resulting in less head injuries.

JanNiklasFeld:
> > @JanNiklasFeld What/why did you change the spline in the engine?
> 
> I think it moves the head at a different point, resulting in less head injuries.

Yes, that was the reason.

Flova:
We now solved the standing up issues

PR #423, opened 2024-04-23
added readme as instruction for extrinsic calibration

00lake00:
# Summary

## Checklist

- [ ] Run `colcon build`
- [x] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


timonegk:
I think you should add in the beginning that roll pitch and yaw of the camera and the imu can be changed and what the difference is. You should also explain what a camera coordinate system is or add a link to the corresponding REP.

jaagut (approved):

00lake00:
@timonegk so fine jetzt?

timonegk:
The image of a right handed coordinate system also shows the left handed system which light be confusing. What I meant was simply linking to https://www.ros.org/reps/rep-0103.html#coordinate-frame-conventions where right-handedness and the convention for optical frames are defined.

bitbots_misc/bitbots_extrinsic_calibration/README.md
```diff
@@ -0,0 +1,39 @@
+# How to: Extrinsic Calibration
+
+In order to adjust the calibration of the visualization you can change the roll (`offset_x`), pitch (`offset_y`) and yaw (`offset_z`) of the camera and the IMU.
```
Flova, line 3: Maybe add a bit of motivation/why we need this.

bitbots_misc/bitbots_extrinsic_calibration/README.md
```diff
@@ -0,0 +1,39 @@
...
+The camera parameters change the camera direction and the IMU parameters change the orientation of the body.
+
+1. Start teamplayer.
```
Flova, line 6: Maybe state that all of them should be started in their own terminal

bitbots_misc/bitbots_extrinsic_calibration/README.md
```diff
@@ -0,0 +1,39 @@
...
+## IMU Parameters
+
+* change if the lines are more far away on one side / to the front
+* has a right-handed coordinate system
+
+![right-handed coordinate system](README_images/right_handed_coordinate_system.png)
+
+## Camera Parameters
+
+* change if the lines are not aligned equally on both sides / front
```
Flova, line 36: Maybe explain that this is the case due to the rotation of the head which lead to an error in the cameras frame of reference being present in all directions.

bitbots_misc/bitbots_extrinsic_calibration/README.md
```diff
@@ -0,0 +1,39 @@
...
+5. Open config file in **bitbots_misc &rarr; bitbots_extrinsic_calibration &rarr; config**
+
+6. Open rqt and navigate to **Plugins &rarr; Configurations &rarr; Dynamic Reconfigure** where you can configure the parameters.
+
+If you change the calibration first change all parameters to `0.0`.
+Then start with the adjustment of the IMU parameters.
+
+## IMU Parameters
+
+* change if the lines are more far away on one side / to the front
```
Flova, line 29: What parameter should be charged if they are unequally distanced to the right/left or to the front / sides.

Flova: Also state that this is an interactive process and we might need to do a few alternating steps of imu and camera calibration to get a good solution.

timonegk (approved):

PR #444, opened 2024-05-06
Time independet ema for path planning smoothing

Flova:
# Summary
Currently, the path planning exponential smoothing is dependent on the update frequency of the path planning. But this can be formulated in a time independent discrete way by using the _time constant tau_ (see https://en.wikipedia.org/wiki/Exponential_smoothing#Time_constant). It described the amount of time a unit impulse takes to decay to 1 - 1/e which is roughly 63%. Using this, we can calculate a smoothing factor alpha for arbitrary time deltas instead of using a fixed one. This accounts for changed rates as well as fluctuations in the timers rate. 

Here you can see two impulse responses with different discretization, but same tau (= 0.39):
image

Note: k and alpha are used as synonyms here for legacy reasons.

## Proposed changes
- Fix typo
- Add flexible time delta
- Use tau instead of k/alpha
- Use ROS 1 parameter value for smoothing. ROS 1 used a slower update frequency for the controller, but with this we can translate the smoothing parameter to the new 20Hz update rate (we need to check that on the real robot!)

## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

Idea: We can add this also to the IMU filter!


SammyRamone (approved):
lgtm

Flova:
We did some parameter tuning, but further testing is required

bitbots_navigation/bitbots_path_planning/config/path_planning.yaml
```diff
@@ -17,13 +17,22 @@ bitbots_path_planning:
...
       max_rotation_vel: 0.4
-      max_vel_x: 0.1
+      # Maximum velocity we want to reach in different directions (base_footprint coordinate system)
+      max_vel_x: 0.15
       min_vel_x: -0.1
-      max_vel_y: 0.03
-      smoothing_k: 0.05
+      max_vel_y: 0.05
+      # This is the time constant of the exponential smoothing filter. The higher the value, the more smoothing is applied.
+      # It denotes the time after which only 63% (1-1/e) of the original signal is left. It is update rate independent!
```
jaagut, line 29: Original signal refers to the latest value and not the last_cmd_vel, right? Maybe use a different description than `original`…

bitbots_navigation/bitbots_localization/config/config.yaml
```diff
@@ -26,12 +26,12 @@ bitbots_localization:
         starting_multiplier: 0.05
         starting_steps_with_higher_diffusion: 40
       drift:
-        distance_to_direction: 2.0
+        distance_to_direction: 0.2
```
jaagut, line 29: Where do these values come from?

Flova: We tuned the localization on the real robot and these seem to be reasonable values. We still need to perform some testing, but for now they narrow the particle cloud quite a bit while still tracking the robot. This allows for a more confident localization, but puts more trust into the odometry and vision.

bitbots_navigation/bitbots_path_planning/config/path_planning.yaml
Flova: Original signal a quite general term in this regard and this indeed refers to the output of the controller and does not consider cmd vel messages from other nodes.

Issue #439, opened 2024-05-02
Use generate_parameter_library in dynup


timonegk:

jaagut:
Done with #445

PR #445, opened 2024-05-07
Dynamic reconfigure dynup

val-ba:
# Summary

We want to use dynamic reconfigure for dynup
## Proposed changes

We want to implement the picknik library for that
## Related issues

[#439](https://github.com/bit-bots/bitbots_main/issues/439)
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
There are a few style chest failing

jaagut:
Also, it looks like, there are a few German notes and questions in the code. Should they be answered by reviews?

val-ba:
> Also, it looks like, there are a few German notes and questions in the code. Should they be answered by reviews?

whoops, I talked to Sebastian and while talking I made notes and forgot about it. I will remove them.

val-ba:
> There are a few style chest failing

How do I change that?

Flova:
> > There are a few style chest failing
> 
> How do I change that?

Run make format in the repo

bitbots_motion/bitbots_dynup/config/dynup_config.yaml
```diff
@@ -0,0 +1,379 @@
+bitbots_dynup:
+  engine: 
+    engine_rate: {
+      type: int,
+      description: "",
```
Flova, line 5: Maybe add descriptions to all params

bitbots_motion/bitbots_dynup/config/dynup_robot.yaml
```diff
@@ -1,60 +1,66 @@
 dynup:
```
Flova, line 1: You have default parameters in the parameter schema and this config file. Try to use only one of them. This sim config file is nessesary tho.

bitbots_motion/bitbots_dynup/include/bitbots_dynup/dynup_stabilizer.hpp
```diff
@@ -11,16 +11,16 @@
...
-
 namespace bitbots_dynup {
 
 class Stabilizer : public bitbots_splines::AbstractStabilizer<DynupResponse> {
  public:
   explicit Stabilizer(std::string ns);
   DynupResponse stabilize(const DynupResponse &response, const rclcpp::Duration &dt) override;
   void setRSoleToTrunk(geometry_msgs::msg::TransformStamped r_sole_to_trunk);
-  void setParams(std::map<std::string, rclcpp::Parameter> params);
+  void setParams(bitbots_dynup::Params params);
```
Flova, line 23: Why are all params passed to the stabilizer and only the engine ones to the engine?

bitbots_motion/bitbots_dynup/src/dynup_ik.cpp
```diff
@@ -30,7 +30,8 @@ void DynupIK::reset() {
 
 void DynupIK::setDirection(std::string direction) { direction_ = direction; }
 
-bitbots_splines::JointGoals DynupIK::calculate(const DynupResponse &ik_goals) {
+bitbots_splines::JointGoals DynupIK::calculate(
+    const DynupResponse &ik_goals) {  // What is DynupResponse? liegt in dynup_utils.h und ist ein struct
```
Flova, line 34: I don't understand that question and answer

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -3,118 +3,78 @@
...
-DynupNode::DynupNode(const std::string &ns, std::vector<rclcpp::Parameter> parameters)
-    : Node(ns + "dynup", rclcpp::NodeOptions()
-                             .allow_undeclared_parameters(true)
-                             .parameter_overrides(parameters)
-                             .automatically_declare_parameters_from_overrides(true)),
+DynupNode::DynupNode(
+    const std::string &ns,
+    std::vector<rclcpp::Parameter> parameters)  // parameters werden im launch script übergeben, aber hierüber kriegt er
+                                                // auch die yaml, also gucken, dass das vlt anders
+    : Node(ns + "dynup", rclcpp::NodeOptions()),  // TODO: vlt. rausnehmen?
```
Flova, line 10: No! Needed for python api

Flova:
Otherwise LGTM

bitbots_motion/bitbots_dynup/include/bitbots_dynup/dynup_stabilizer.hpp
val-ba: no reason. It will be changed

val-ba: The yaml also has some wrong indentations, semantically at least.

bitbots_motion/bitbots_dynup/include/bitbots_dynup/dynup_stabilizer.hpp
```diff
@@ -20,7 +20,7 @@ class Stabilizer : public bitbots_splines::AbstractStabilizer<DynupResponse> {
   explicit Stabilizer(std::string ns);
   DynupResponse stabilize(const DynupResponse &response, const rclcpp::Duration &dt) override;
   void setRSoleToTrunk(geometry_msgs::msg::TransformStamped r_sole_to_trunk);
-  void setParams(bitbots_dynup::Params params);
```
Flova: Maybe pass stabilizer

bitbots_motion/bitbots_dynup/src/dynup_ik.cpp
val-ba: old comment, forgot to remove it

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: okay!

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
```diff
@@ -1,75 +1,49 @@
...
-    # Descend
-    descend_time: 0.25
+    # Rise
+      rise:
+        rise_time: 0.4 
 
     # Stabilier
     stabilizing: True
     minimal_displacement: False
-    stable_threshold: 0.05
```
Flova, line 65: Is the internation/hierarchy correct here? Isn't it missing the stabilizer namespace?

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -130,7 +87,10 @@ DynupNode::DynupNode(const std::string &ns, std::vector<rclcpp::Parameter> param
   engine_.init(shoulder_origin.position.y, shoulder_origin.position.z);
   ik_.init(kinematic_model_);
 
-  callback_handle_ = this->add_on_set_parameters_callback(std::bind(&DynupNode::onSetParameters, this, _1));
+  callback_handle_ = this->add_on_set_parameters_callback(
```
Flova, line 90: Why do we have an on set parameters callback? I didn't notice this last time. But normally the generate parameter library updates via the parameter listener at the beginning of each execution step/when the dynup is called.

Flova: You seem to do that. But why are you also using that callback?

val-ba: Because the old dynup used this callback and I used it at first to keep the old structure. I don't know when the callback handle is called and I thought maybe it is important to call it at the same time. But it should probably not make a difference, when we check whether the parameters have been updated in the dynup loop. Should it be removed?

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
val-ba: Looking at the old file, it looked like this without another parameter. one level above. Does it need to reflect the yaml which the parameter library uses? Because it is true that there is stabilizer one hierarchy level above.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
Flova: I think it should be removed as this is mixing the old and new style of reconfiguration, which might lead to issues.

I would rather check if the parameters are old when we get a goal request and then call the callback function after getting the params instead of registering the Ros callback.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -223,6 +179,10 @@ void DynupNode::execute(const std::shared_ptr<DynupGoalHandle> goal_handle) {
   last_ros_update_time_ = 0;
   start_time_ = this->get_clock()->now().seconds();
 
+  params_ = param_listener_.get_params();
```
Flova, line 182: Maybe only update the params if they are outdated.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -223,6 +179,10 @@ void DynupNode::execute(const std::shared_ptr<DynupGoalHandle> goal_handle) {
...
+  // Copy engine params to engine
+  engine_.setParams(params_.engine);
```
Flova, line 184: Calling the on Parameter Change function should be the more general solution here

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -302,6 +262,7 @@ double DynupNode::getTimeDelta() {
 }
 
 void DynupNode::loopEngine(int loop_rate, std::shared_ptr<DynupGoalHandle> goal_handle) {
+  params_ = param_listener_.get_params();
```
Flova, line 265: What is the purpose of this get params here? They params should be updated when the goal is received. Also they are not passed to the engine etc..

Flova (changes requested):

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: Why? I think it does not change much or does it diminish the performance?

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: i know, I just noticed that aswell.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
Flova: I also don't think it's that bad, but the examples all do it

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: I noticed that the walk_node also has one, but it does not seem to be used. Then I will remove it.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: Okay, will do.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
Flova: Whz has this been resolved? You don't call the function and setParams is manually called for the engine only.

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
Flova: You added the `stabilizer` namespace in the template/default yaml. Not using it here leads to all stabilizer related parameters not being updated to the override, as this line here just defines a value for a parameter that doesn't exist (due to the design of the ros2 params you don't get an error when setting undeclared parameters in the config, as they could be declared later during runtime.

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
```diff
@@ -78,8 +52,6 @@ dynup_pid_trunk_fused_roll:
     i_clamp: 0.0
     i_clamp_min: 0.0
     i_clamp_max: 0.0
-    antiwindup: True
```
16stelter, line 81: Is there a reason for removing these parameters? Anti-Windup is useful to reduce pid overshoot

bitbots_motion/bitbots_dynup/src/dynup_engine.cpp
```diff
@@ -749,35 +732,27 @@ double DynupEngine::getDuration() const { return duration_; }
 
 /*Calculates if we are at a point of the animation where stabilizing should be applied. */  // TODO: make this nice
 bool DynupEngine::isStabilizingNeeded() {
```
16stelter, line 734: This function could be further refactored, but it works like this. In any case, remove the todo from the comment.

bitbots_motion/bitbots_dynup/src/dynup_engine.cpp
```diff
@@ -749,35 +732,27 @@ double DynupEngine::getDuration() const { return duration_; }
...
+                                           params_.dynup_front.time_foot_close + params_.dynup_front.time_hands_front +
+                                           params_.dynup_front.time_foot_ground_front +
+                                           params_.dynup_front.time_torso_45 + params_.dynup_front.time_to_squat) ||
+          (direction_ == 0 && time_ >= params_.dynup_back.time_legs_close + params_.dynup_back.time_foot_ground_back +
+                                           params_.dynup_back.time_full_squat_hands +
+                                           params_.dynup_back.time_full_squat_legs) ||
           (direction_ == 2) || (direction_ == 3));
 }
 
 bool DynupEngine::isHeadZero() {
```
16stelter, line 745: Could be refactored, see above.

Flova:
See my answers

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
```diff
@@ -4,117 +4,74 @@ namespace bitbots_dynup {
...
   std::string check_kinematic_parameters;
-  if (!this->get_parameter("robot_description_kinematics.LeftLeg.kinematics_solver", check_kinematic_parameters)) {
+  if (!moveit_node->get_parameter("robot_description_kinematics.LeftLeg.kinematics_solver",
+                                  check_kinematic_parameters)) {
     auto parameters_client = std::make_shared<rclcpp::SyncParametersClient>(this, "/move_group");
     while (!parameters_client->wait_for_service(1s)) {
       if (!rclcpp::ok()) {
         RCLCPP_ERROR(this->get_logger(), "Interrupted while waiting for the service. Exiting.");
-        rclcpp::shutdown();
+        break;
```
16stelter, line 35: This should not be break, if rclcpp has been shut down, the rest of the node should not be attempted to be executed.

bitbots_motion/bitbots_dynup/src/dynup_engine.cpp
```diff
@@ -51,66 +51,51 @@ void DynupEngine::publishDebug() {
   msg.time = time_;
   msg.stabilization_active = isStabilizingNeeded();
   if (direction_ == 1) {
-    if (time_ < params_["time_hands_side"].get_value<double>()) {
+    if (time_ < params_.dynup_front.time_hands_side) {
       msg.state_number = 0;
```
16stelter, line 55: This is a really not so great remnant from my thesis when I was in  a rush and definitely should be improved.

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
```diff
@@ -1,75 +1,49 @@
...
 
     # Stabilier
     stabilizing: True
     minimal_displacement: False
-    stable_threshold: 0.05
     stable_duration: -1
     stabilization_timeout: 1.5
 
-    # Visualizer
-    spline_smoothness: 100
```
16stelter, line 70: these parameters exist in dynup_config.yaml, but have been removed here. Why?

bitbots_motion/bitbots_dynup/config/dynup_config.yaml
```diff
@@ -0,0 +1,379 @@
...
+        type: int,
+        description: "",
+        default_value: 100,
+      }
+      display_debug: {
+        type: bool,
+        description: "",
+        default_value: true,
+      }
+
```
16stelter, line 379: Parameters of the pid controllers are missing here, would be good to make them dynamic as well

16stelter (changes requested):
Few minor issues, but mostly okay. Make sure this is tested extensively on the robot before merging.

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
val-ba: Because they do not differ from the default values in the dynup_config. If they are different, they are specified. We used the default values from the robot_config and looked at the values from the sim_config and only wrote the ones down, which are different.

bitbots_motion/bitbots_dynup/config/dynup_sim.yaml
val-ba: OOh, okay. I will change that.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: I used this code from the walk node, where it breaks. rclcpp::shutdown() has been removed.

bitbots_motion/bitbots_dynup/src/dynup_node.cpp
val-ba: Then I misunderstood what you meant. What do you mean by parameter change function?

Flova (approved):

PR #451, opened 2024-05-08
No share dsd code

Flova:
# Summary
Currently we explicitly copy the dsd related files into the share directory during install. But this is not necessary. As they are part of the python module they are installed the same way as the regular python code during install. This also means thy are probably symlinked etc.. To get the install path we can just import them the normal way and get their `__path__` attribute. 

## Proposed changes

- Import all dsds in the described manner

Todo:
- [x] Behavior 
- [x] Localization 
- [x] HCM

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


Flova:
We might also want to directly pass the module to the dsd initialization instead of just the path, so we don't need to import twice. But I am not sure it this is really beneficial nor how to do it properly. 

We could also accept a module or string on the dsd side and extract the path there so the implementation details are less exposed.

Any thoughts @timonegk

timonegk (approved):

timonegk:
> We might also want to directly pass the module to the dsd initialization instead of just the path, so we don't need to import twice. But I am not sure it this is really beneficial nor how to do it properly. 
> 
> We could also accept a module or string on the dsd side and extract the path there so the implementation details are less exposed.
> 
> Any thoughts @timonegk 

Yes that sounds like it would make the usage more simply. But double import is not a problem because there is not really any overhead if you import a module that is already imported.

Flova:
Todo: fix tests

PR #457, opened 2024-05-16
Efficient python tf buffer

Flova:
# Summary

Instead of having a custom tf listener this PR replaces the whole buffer. Python code is only involved if we actually do a transform.

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
This needs some cleanup and better handling of tf exceptions, but it generally works!

Flova:
Any comments @timonegk ?

bitbots_misc/bitbots_tf_buffer/README.md
```diff
@@ -1,9 +1,9 @@
-# bitbots_tf_listener
+# bitbots_tf_buffer
 
 This is a drop-in replacement for `tf2_ros.TransformListener` in Python. It implements the callback for the
```
jaagut: Shure it's not the Buffer?

PR #458, opened 2024-05-16
Non root devcontainer

texhnolyze:
# Summary
This fixes the issue, that when utilizing the devcontainer but at the same time interacting with the repository from the host we run into issues, as files within the repo have the wrong `UID/GID` from within the container and are not readable, because we use the `root` user in the container.

## Proposed changes
- update Dockerfile to include a `bitbots` user with `uid=1000, gid=1000`
- make the user configurable both in name and `UID`, `GID` by utilizing `Dockerfile` `ARG`
- allow `bitbots` user to run privileged commands by adjusting `/etc/sudoers`
- remove the `users` group in the container

With these changes, the devcontainer will use the `bitbots` user internally, which is initially setup with `uid=1000, gid=1000`, but vscode will switch the `uid` and `gid` dynamically to the one of the host user (see https://github.com/microsoft/vscode-remote-release/issues/1155).

## Checklist

- [x] Test on your machine
- [x] Test on CL

### DIFF
--- a/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/misc_capsule.py
+++ b/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/misc_capsule.py
@@ -1,6 +1,6 @@
 from typing import Literal, Optional, TypeAlias
 
-from bitbots_utils.utils import get_parameters_from_other_node
+from bitbots_utils.utils import RobotNotConfiguredError, get_parameters_from_other_node
 from rclpy.duration import Duration
 from std_msgs.msg import Bool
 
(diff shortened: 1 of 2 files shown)


jaagut:
When testing with a fresh container, i get this:
```shell
 🐋  ~/colcon_ws/src/bitbots_main  gc --allow-empty -m "Foo"
ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
clang-format.........................................(no files to check)Skipped
cppcheck.............................................(no files to check)Skipped
cmake-format.........................................(no files to check)Skipped
cmake-lint...........................................(no files to check)Skipped
check for merge conflicts............................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check xml............................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
detect private key...................................(no files to check)Skipped
fatal: cannot lock ref 'HEAD': Unable to create '/home/bitbots/colcon_ws/src/bitbots_main/.git/refs/heads/feature/non-root-dev-container.lock': Permission denied
```

texhnolyze:
Interesting, I am unable to reproduce the issue. I would assume that the lock file can either not be created due to some issues with permissions or because you have some other git process running already, which has already created the file.
Could you tell me which user/group the files under `/home/bitbots/colcon_ws/src/bitbots_main/.git/refs` belong to?
For me, it is `bitbots:bitbots`, but if you are on an Ubuntu host for example it would most likely be `bitbots:users` as we are using the same primary group as the host user account.
But that should still not be an issue.

Issue #313, opened 2022-12-07
HCM callback groups


SammyRamone:
The HCM has a lot of callbacks. To achieve a decent performance, these should be executed in parallel. This is unfortunately not the default. See also https://docs.ros.org/en/rolling/How-To-Guides/Using-callback-groups.html

## Expected behavior
All callbacks for sensordata and joint goals is executed in parallel

## Current behavior
All callbacks are executed after each other.

## Steps to Reproduce

1.
2.
3.
...

## Context (Environment)

- [ ] RViz
- [ ] Simulator
- [ ] Robot
- [ ] Local

PR #460, opened 2024-05-16
feat(hcm): setup callback groups for high hz subscriptions

texhnolyze:
using a `MutuallyExclusiveCallbackGroup` for each one, so that they do not block each other but each subscription's callback is still called in order of received messages.

See: https://docs.ros.org/en/rolling/How-To-Guides/Using-callback-groups.html

Fixes: #313

## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_motion/bitbots_hcm/src/hcm.cpp
+++ b/bitbots_motion/bitbots_hcm/src/hcm.cpp
@@ -62,13 +62,21 @@ class HCM_CPP : public rclcpp::Node {
 
     // Create subscriber for high frequency sensor data
     joint_state_sub_ = this->create_subscription<sensor_msgs::msg::JointState>(
-        "joint_states", 1, std::bind(&HCM_CPP::joint_state_callback, this, _1));
+        "joint_states", 1, std::bind(&HCM_CPP::joint_state_callback, this, _1), high_hz_sub_options());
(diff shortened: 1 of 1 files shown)


texhnolyze:
I'd like to test this on a robot, before finishing up. But it can already be reviewed.

timonegk (approved):
Code-wise this looks fine but I cannot judge the performance increase.

texhnolyze:
I took a look at the `EventsExecutor` implementation and determined, that it is single threaded and utilizing a single FIFO queue, no matter the callback groups.
To actually utilize parallel execution, we would need to either need to utilize an `MultiThreadedExecutor` with a multiple threads or assign different callback groups to separate instances of `EventsExecutor` utilizing the `->add_callback_group` method.

In the way it was implemented right now, it will most likely not have any effect on the execution of callbacks in the HCM, but will also not break anything.

Until some clarification and benchmarking, I would put this on hold.
@Flova did we even still have issues with speed of `joint_states` and `imu/data` in the HCM in the last months?

Flova:
I don't think so tbh. as discussed in person. (Just writing it here as a protocol for the future.

PR #455, opened 2024-05-14
Basler camera diagnostics

Flova:
# Summary
Adds diagnostics to the basler camera driver

## Proposed changes
- Fix terminal spam
- Add diagnostics for errors and camera cover to the basler driver
- Make diagnostics aggregator global (it was part of lowlevel before)

## Related issues
Closes #435
Closes #454  

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [x] Test on the robot
- [x] Create issues for future work
- [x] Triage this PR and label it

PR #463, opened 2024-05-17
vision: remove camera cap check and unused speak capability

### LINKED ISSUE #462: Vision often wrongly says image too dark/ camera cap on

jaagut:
# Summary

Removes redundant camera cap check in vision

## Proposed changes

As the camera cover/cap check has been implemented in the camera driver with proper diagnostics, this check in the vision became redundant.
This PR removes the vision check and the speak capability, as it has no use anymore.

## Related issues

Closes #462
#455

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
I still think that the audio warning is nice. Even if the functionality is now located in the basler driver. We should debug either way why the audio is not played during startup.

jaagut:
If I remember correctly, we now have a simpler function in misc to speak. We should use that one instead. And I agree, we should speak this warning. However, I don't think it should be located in this vision anymore.

Flova:
> function in misc

I don't know what you mean. There is a simple factory for the speak messages, but it is barely needed and only python (the basler driver is cpp).

Flova (approved):

jaagut:
Yes, that's the one I meant. And yes, its only for python nodes...

My preference for which node should say the warning:

1. Gameready-script
2. Something in Misc
3. Vision
4. Basler driver

Flova:
I think integrating it into the basler would be fitting as we have a check there either way. But I just had the following idea:

Diagnostic state to tts! We just say (some) if the arbitrary diagnostic states. Therefore the robot tells us if the camera is overheating, the cap is on there or if the motors are stale/overloaded.

PR #465, opened 2024-05-20
Feature/technical challenge vision

confusedlama:
# Summary
Adds a simple dynamically configurable vision package that detects obstacles for the obstacle avoidance technical challenge.

## Proposed changes
-New Package "bitbots_misc/bitbots_technical_challenge_vision"
-Detect red and blue obstacles by isolating color, getting contours and finding the bounding rectangle for each contour
-Dynamically adjust color ranges and min size of bounding box in rqt
-Dynamically toggle debug mode in rqt
-Publish annotations as RobotArray messages

## Questions
-Are the team numbers relevant?
-Should RobotArray messages be published in "ROS_robot_msg_topic" instead?


bitbots_misc/bitbots_technical_challenge_vision/bitbots_technical_challenge_vision/bitbots_technical_challenge_vision.py
```diff
@@ -168,6 +169,11 @@ def image_callback(self, msg: Image):
             self._debug_clrmp_pub_blue.publish(clrmp_blue_msg)
             self._debug_clrmp_pub_red.publish(clrmp_red_msg)
 
+    def timer_callback(self):
```
Flova, line 172: You can just check during the image callback if I don't miss anything.

confusedlama:
Should RobotArray messages be published in "ROS_robot_msg_topic" instead of a new topic? @Flova

Flova:
Use the same topic as the vision. Then you don't need to change anything downstream. The name you wrote is the name of a parameter where the topic is located (I don't like this pattern, but it is present in the vision). Just use the parameter value from the config. Should be something like `/robots_in_image`>

Flova:
I read the rules and the behavior should be pretty straightforward too. It just needs to initialize the localization at the correct point (not strictly needed, as we are starting at the sidelines, but we can reduce the initial localization time by telling the robot where it is), after that it needs to send a goal pose at the other side of the field (clearing the side line). All of that should start when we press a button. I would not use the DSD for that as it can be realized in a very simple script.

The robot should not leave the area of the field where the challenge takes place. One hack to achieve this without modifying the path planning would be to publish a large number of robots around the field in an fixed interval. This should result in the robot not leaving the area. The clean way would be to draw it properly onto the nav map in the path planning.

confusedlama:
I think we should make a separate issue for the behaviour and merge this branch if you consider the vision to be complete.

bitbots_misc/bitbots_technical_challenge_vision/bitbots_technical_challenge_vision/bitbots_technical_challenge_vision.py
```diff
@@ -0,0 +1,187 @@
...
+            def annotate(x, y, h, w, c):
+                return None
+
+        for cnt in blue_contours:
+            x, y, h, w = cv2.boundingRect(cnt)
+            if min(h, w) >= arg.min_size and max(h, w) <= arg.max_size:
+                # draw bb on debug img
+                annotate(x, y, h, w, (255, 0, 0))
+
+                # TODO I think 1 is for the blue team?
```
Flova, line 113: You can/should use the constant in the message for that.

Flova:
Maybe we could rename it to "bitbots_obstacle_challange_vision"

Flova (approved):
otherwise lgtm

PR #466, opened 2024-05-23
Remove scaling for sole to base

val-ba:
# Summary

It may not be useful to scale the transform from one foot to the other with the same factors as the support sole to the base.
Additionally, the transform from support foot to base might not have to be scaled at all.
## Proposed changes

Remove the scaling factor from sole to base.
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
Might be superseded by #469

Issue #424, opened 2024-04-28
Only do velocity estimation for close balls in ball filter


Flova:
# Summary
Projection noise leads to false velocity estimates for far away balls. This leads to the world model ball rapidly leaving the field, resulting in a lost ball / wrong ball position. 

## Expected behavior
- We should account for increased noise for far a way balls in the velocity estimation.

## Current behavior
- We assume a wrong vel which fucks up the estimated ball position.

## Steps to Reproduce
- See rosbag from amy in the second half of the final of the german open 2024

## Possible Solution
- Ignore velocity if the object if further away than a threshold
- Properly handle the increased process noise

PR #468, opened 2024-05-27
Add dynamic reconfigre to ball filter

val-ba:
# Summary

We want to tune the velocity estimation of the ball filter #424.
It is useful to be able to dynamically reconfigure the parameters.
## Proposed changes

Use https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
+++ b/bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
@@ -1,6 +1,5 @@
 #! /usr/bin/env python3
 import math
-from copy import deepcopy
 from typing import Tuple, Union
 
 import numpy as np
@@ -19,6 +18,7 @@ from std_msgs.msg import Header
 from std_srvs.srv import Trigger
 from tf2_geometry_msgs import PointStamped
 
+from bitbots_ball_filter.ball_filter_parameters import BallFilterParameters as Parameters
 from bitbots_msgs.msg import PoseWithCertaintyStamped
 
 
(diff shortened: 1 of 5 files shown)


bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
```diff
@@ -0,0 +1,94 @@
...
+  filter_frame: {
+    type: string,
+    default_value: 'map',
+    description: 'Frame to filter the ball in',
+    validation: {
+      one_of<>: ['map', 'odom']
+    }
+  }
+
+  filter_rate: {
```
Flova, line 36: I think this param is read only, as it is not changeable during runtime

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
```diff
@@ -0,0 +1,95 @@
...
+  ball_movement_publish_topic: {
+    type: string,
+    default_value: 'ball_relative_movement',
+    description: 'Topic to publish the filtered ball movement'
+  }
+  ball_publish_topic: {
+    type: string,
+    default_value: 'ball_relative_filtered',
+    description: 'Topic to publish the filtered ball'
+  }
```
jaagut, line 21: These topic parameters should also either be read-only or react to changes.

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
```diff
@@ -0,0 +1,95 @@
...
+  ball_filter_reset_service_name: {
+    type: string,
+    default_value: 'ball_filter_reset',
+    description: 'Service to reset the ball filter'
+  }
```
jaagut, line 26: Same for this one

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  process_noise_variance: {
+    type: double,
+    default_value: 0.001,
+    description: 'Noise which is added to the estimated position without new measurements',
+    validation: {
+      bounds<>: [0, 1]
+    }
+  }
```
jaagut, line 62: This should also be read-only or reactive

jaagut (changes requested):

jaagut:
@Flova how do we handle the autogenerated `*parameters.py` files in other places? I remember in some packages, we also generate these files and I have manually avoided committing these.
I am not sure, but I believe these files are actually generated inside the build/ or install/ directories, but due to symlinking they appear in the src/ directory where they can accidentally be committed. We could just git-ignore `*parameters.py` files, but that's not a 100% clean solution.

Flova:
This is no easy issue. You can commit them, as they are a code gen result, so you get auto completion etc before building. But I agree that this is kind of ugly. The generate parameter library indeed generates them in the src dir when used with setup.py. This would be the first python package to utilize this. Cpp code gen gets directly put into the build folders. I would just add it to the gitignore tbh.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -297,12 +303,14 @@ def publish_data(self, state_vec: np.array, cov_mat: np.array) -> None:
...
     node = BallFilter()
     # Number of executor threads is the number of MutiallyExclusiveCallbackGroups + 1 thread for the executor
-    ex = MultiThreadedExecutor(num_threads=3)
-    ex.add_node(node)
+    # ex = MultiThreadedExecutor(num_threads=3)
+    # ex.add_node(node)
     try:
-        ex.spin()
+        # ex.spin()
+        rclpy.spin(node)
```
Flova, line 313: Why single threaded executor again?

jaagut:
@Flova ahh, I got this already in the template repo!

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
val-ba: What do you mean by reactive?

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
val-ba: woops, I tried fixing errors and did that, i will change it back

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter_parameters.py
```diff
@@ -0,0 +1,342 @@
+# flake8: noqa
+
+# auto-generated DO NOT EDIT
```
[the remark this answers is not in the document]
val-ba: I removed it.

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
val-ba: Ah okay, understood, I do not use them but only in the beginning when initializing the publisher and subscriber. I thought there was an extra field called "reactive" or something like that.

bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
val-ba: they are reactive now.

jaagut (approved):

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -140,28 +132,19 @@ def _get_transform(
...
         if frame is None:
-            frame = self.filter_frame
+            frame = self.config.filter_frame
 
         point_stamped = PointStamped()
         point_stamped.header = header
         point_stamped.point = point
 
         try:
-            return self.tf_buffer.transform(point_stamped, frame, timeout=Duration(seconds=timeout))
```
Flova, line 160: This timeout is 0 all the time, because 0.3 gets casted to an int

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -173,17 +156,25 @@ def filter_step(self) -> None:
         if self.param_listener.is_old(self.config):
             self.param_listener.refresh_dynamic_parameters()
             self.config = self.param_listener.get_params()
-            self.set_parameters()
+            self.filter_reset_duration = Duration(seconds=self.config.filter_reset_time)
+            self.velocity_factor = (1 - self.config.velocity_reduction) ** (self.filter_time_step)
+            self.kf.Q = Q_discrete_white_noise(
```
Flova, line 161: This seems to be initialized only if dynamic reconfigure is triggered at least once. Maybe put these on a function and call it in the beginning when you pull the params as well as here.

Flova (changes requested):

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
val-ba: changed it to using nanoseconds.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
val-ba: I need to wait until the variables are created, but I can call it afterswards.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -60,6 +60,7 @@ def __init__(self) -> None:
         self.filter_reset_duration = Duration(seconds=self.config.filter_reset_time)
         self.logger.info(f"Using frame '{self.config.filter_frame}' for ball filtering")
 
+        self.update_params()
```
Flova, line 63: Now a bunch of params are updated twice. Remove their updates from the constructor, as it is happening in the update params function. Also the config is retrieved twice.

Flova (changes requested):

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
val-ba: As far as I can see it, it is only the filter_reset_duration. I can set it to None first?

Flova: You can just set it in the function

Flova (approved):

PR #472, opened 2024-05-30
Rework field config 

### LINKED ISSUE #447: Unified field configuration
# Summary

Currently the localization and behavior/parameter blackboard have different configs for the field dimension. This leads to cases where the localization looks good, but the behavior tells the robot to leave the field as we assume that the field is larger than it is. 

It would also be nice to have multiple parameter sets so we can swap between them without reverting multiple related params.

The field configuration is not obvious.

Add more information regarding the field to the parameter blackboard. We need to see how the map is handled, as it is quite closely related to the localization.

Keep it as it is now.

Flova:
To rework the field config, I moved the parameter blackboard into its own package. All field related config including the maps has been moved there. Field specific parameters are now able to change based on the selected field. I also removed some unnecessary parameters from the localization and clarified some namings, primary in the localization and behavior.

# Summary
- Don't define things like the field size at multiple places
- Cleanup field related things
- New package for the parameter blackboard

## Related issues
Closes #447 

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


bitbots_misc/bitbots_utils/src/utils.cpp
```diff
@@ -61,4 +61,39 @@ void wait_for_tf(const rclcpp::Logger &logger, std::shared_ptr<rclcpp::Clock> cl
...
+  for (size_t i = 0; i < parameter_names.size(); i++) {
+    // Convert to parameter message fist, so we have less hassle with the params type template
+    auto parameter = rcl_interfaces::msg::Parameter();
+    parameter.name = parameter_names[i];
+    parameter.value = response->values[i];
+    results[parameter_names[i]] = rclcpp::Parameter::from_parameter_msg(parameter);
+  }
+
+  return results;
+}
```
jaagut, line 98: This function looks fine for me. I was wondering, what happens, when a non-existing parameter gets requested.

bitbots_navigation/bitbots_localization/include/bitbots_localization/localization.hpp
```diff
@@ -71,9 +72,9 @@ using namespace std::placeholders;
  * @class Localization
  * @brief Includes the ROS interface, configuration and main loop of the Bit-Bots RoboCup localization.
  */
-class Localization : public rclcpp::Node {
+class Localization {
  public:
-  explicit Localization();
+  explicit Localization(rclcpp::Node::SharedPtr node);
```
jaagut, line 77: What's the difference?

Update: Ahh, is it necessary to share this node with the get parameters from other node service?

bitbots_navigation/bitbots_localization/include/bitbots_localization/localization.hpp
```diff
@@ -182,12 +190,9 @@ class Localization : public rclcpp::Node {
   std::shared_ptr<particle_filter::ParticleFilter<RobotState>> robot_pf_;
 
   // Declare initial state distributions
-  std::shared_ptr<RobotStateDistributionStartLeft> robot_state_distribution_start_left_;
-  std::shared_ptr<RobotStateDistributionStartRight> robot_state_distribution_start_right_;
-  std::shared_ptr<RobotStateDistributionRightHalf> robot_state_distribution_right_half_;
-  std::shared_ptr<RobotStateDistributionLeftHalf> robot_state_distribution_left_half_;
-  std::shared_ptr<RobotStateDistributionPosition> robot_state_distribution_position_;
-  std::shared_ptr<RobotStateDistributionPose> robot_state_distribution_pose_;
```
jaagut, line 190: Why did you remove these two?

bitbots_navigation/bitbots_localization/src/ObservationModel.cpp
```diff
@@ -8,12 +8,14 @@ namespace bitbots_localization {
...
-  map_lines_ = map_lines;
-  map_goals_ = map_goals;
-  map_field_boundary_ = map_field_boundary;
-  config_ = config;
+                                                     const bitbots_localization::Params &config,
+                                                     const FieldDimensions &field_dimensions)
+    : particle_filter::ObservationModel<RobotState>(),
+      map_lines_(map_lines),
+      map_goals_(map_goals),
+      map_field_boundary_(map_field_boundary),
```
jaagut, line 16: What are these for? Do we still support those other field features?

bitbots_navigation/bitbots_localization/src/StateDistribution.cpp
```diff
@@ -6,23 +6,6 @@
...
-  max_x_ = field_size.first / 2.0 - initial_robot_pose.first;
-  max_y_ = field_size.second / 2.0 - initial_robot_pose.second;
-}
-
-const RobotState RobotStateDistribution::draw() const {
-  return (RobotState(random_number_generator_.getUniform(min_x_, max_x_),
-                     random_number_generator_.getUniform(min_y_, max_y_),
-                     random_number_generator_.getUniform(-M_PI, M_PI)));
-}
-
```
jaagut, line 25: These were unused?

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -5,52 +5,45 @@
...
-                                    rclcpp::Duration::from_nanoseconds(1e9 * 20.0));
-      break;
-    } catch (const tf2::LookupException &ex) {
-      RCLCPP_INFO(this->get_logger(), "Transforms not available, waiting for them... \n %s", ex.what());
-      std::this_thread::sleep_for(std::chrono::milliseconds(100));
-    }
-  }
-  RCLCPP_INFO(this->get_logger(), "Transforms are available now");
+  bitbots_utils::wait_for_tf(node->get_logger(), node->get_clock(), tfBuffer.get(), {config_.ros.base_footprint_frame},
+                             config_.ros.odom_frame);
```
jaagut, line 17: Why did you swap base_footprint and odom?

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -277,8 +272,8 @@ void Localization::reset_filter(int distribution, double x, double y) {
   robot_pf_->setResamplingStrategy(resampling_);
 
   if (distribution == 3) {
-    robot_state_distribution_position_.reset(new RobotStateDistributionPosition(random_number_generator_, x, y));
-    robot_pf_->drawAllFromDistribution(robot_state_distribution_position_);
+    robot_pf_->drawAllFromDistribution(
+        std::make_shared<RobotStateDistributionPosition>(random_number_generator_, x, y));
```
jaagut, line 276: This is a weird change...

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -289,8 +284,8 @@ void Localization::reset_filter(int distribution, double x, double y, double ang
   robot_pf_->setResamplingStrategy(resampling_);
 
   if (distribution == 4) {
-    robot_state_distribution_pose_.reset(new RobotStateDistributionPose(random_number_generator_, x, y, angle));
-    robot_pf_->drawAllFromDistribution(robot_state_distribution_pose_);
+    robot_pf_->drawAllFromDistribution(
+        std::make_shared<RobotStateDistributionPose>(random_number_generator_, x, y, angle));
```
jaagut, line 288: Also weird...

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -566,7 +561,9 @@ void Localization::publish_debug_rating(std::vector<std::pair<double, double>> m
 
 int main(int argc, char *argv[]) {
   rclcpp::init(argc, argv);
-  rclcpp::spin(std::make_shared<bitbots_localization::Localization>());
+  auto node = rclcpp::Node::make_shared("bitbots_localization");
+  [[maybe_unused]] auto localization = bitbots_localization::Localization(node);
```
jaagut, line 565: [[weired syntax]] xD

bitbots_misc/bitbots_parameter_blackboard/config/fields/german_open_2024/config.yaml
```diff
@@ -0,0 +1,22 @@
...
+        padding: 1.0   # Padding (area outside of the lines) of the field in all directions
+      markings:
+        penalty_area:
+          size:
+            x: 2.0   # The distance from the goal line to the boundary of the penalty area
+        goal_area:
+          size:
+            x: 1.0   # The distance from the goal line to the boundary of the goal area
+      goal:
+        width: 2.6   # The width of the goal (center of one goal post to the other goal post)
```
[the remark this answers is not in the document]
Flova: I thought about that, but I think everybody knows what is meant with width of the goal. Width of the field is a bit more ambiguous in im opinion. Therefore I opted to use a different naming. Also the goal has no x depth yet so it is just one value either way.

bitbots_misc/bitbots_parameter_blackboard/config/fields/labor/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+        x: 5.45
+        y: 3.88
+        padding: 0.2   # Padding (area outside of the lines) of the field in all directions
+      markings:
+        penalty_area:
+          size:
+            x: 0.75   # The distance from the goal line to the boundary of the penalty area
+        goal_area:
+          size:
+            x: 0.75   # The distance from the goal line to the boundary of the goal area
```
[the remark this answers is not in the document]
Flova: Not used in the code yet, so I don't think we should have a parameter for that.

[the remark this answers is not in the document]
Flova: It is currently not utilized in the code. I prefer not having unused parameters.

bitbots_misc/bitbots_parameter_blackboard/launch/parameter_blackboard.launch
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+<launch>
+    <arg name="sim" default="false"/>
+    <arg unless="$(var sim)" name="fieldname" default="labor" description="Loads field settings (labor, webots, gazebo, bangkok, spl, german_open_2024)." />
+    <arg if="$(var sim)" name="fieldname" default="webots" description="Loads field settings (labor, webots, gazebo, bangkok, german_open_2024)." />
+
+    <!-- print bit-bot -->
+    <include file="$(find-pkg-share bitbots_utils)/launch/welcome.launch" />
```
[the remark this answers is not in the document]
Flova: I agree it is a bit hacky. But the blackboard is only launched once in multiple launch files e.g. also motion standalone etc. and adding it here is an easy solution to printing it every time without forgetting it.

bitbots_navigation/bitbots_localization/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(ament_index_cpp REQUIRED)
 find_package(backward_ros REQUIRED)
+find_package(bitbots_utils REQUIRED)
```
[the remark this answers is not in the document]
Flova: I might forgot the parameter blackboard dep here. But utils is needed for the use of the get parameters from other nodes utility function.

bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/world_model_capsule.py
```diff
@@ -67,13 +67,19 @@ def __init__(self, blackboard: "BodyBlackboard"):
...
+                "field.size.y",
+            ],
         )
-        self.field_length: float = parameters["field_length"]
-        self.field_width: float = parameters["field_width"]
-        self.goal_width: float = parameters["goal_width"]
-        self.penalty_area_length: float = parameters["penalty_area_length"]
-        self.goal_area_length: float = parameters["goal_area_length"]
+        self.goal_width: float = parameters["field.goal.width"]
+        self.goal_area_size_x: float = parameters["field.markings.goal_area.size.x"]
```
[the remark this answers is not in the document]
Flova: It is indeed never used. I also thought of removing it and it might make sense to do so

bitbots_misc/bitbots_utils/src/utils.cpp
Flova: Good question, I think it is not returned in the map and accessing it fails. But I didn't test it.

bitbots_navigation/bitbots_localization/include/bitbots_localization/localization.hpp
Flova: Exactly and you can not use the service if your node isn't initiated yet (which it isn't when the constructor is called).

bitbots_navigation/bitbots_localization/include/bitbots_localization/localization.hpp
Flova: Because they are a distributions around a point or pose and it doesn't make sense to create them without knowing the requested point or pose, so they need to be created on demand and there is no benefit in keeping them after sampling during the reinitialization. The sideline distribution as an counter example can be reused indefinitely when we initialize on the sideline, as it is not changing.

bitbots_navigation/bitbots_localization/src/ObservationModel.cpp
Flova: The Localization does support them. But we don't use them. I thought about removing them a few times, but it was not worth the effort for me right now. But I also wanted to be consistent in my changes so I also changed them. 


What the code does is initializing the fields with the arguments during the constructor call.

bitbots_navigation/bitbots_localization/src/StateDistribution.cpp
Flova: Exactly. And I don't know what it would be useful for.

bitbots_navigation/bitbots_localization/src/localization.cpp
Flova: Swap? Also it doesn't matter which direction we check the transforms for. We just check for availability.

bitbots_navigation/bitbots_localization/src/localization.cpp
Flova: As I said earlier it doesn't make sense to keep the state distribution for a specific pose and position for longer that it's sampling process. Therefore I create a distribution around the pose and pass that to the sample function. I need to make it a shared pointer, because that's what the interface of that function expects. I don't keep any references to it because it is not useful after the sampling. This is not the case for e.g. the sidelines distribution, which is created once and reused for all sideline reinitializations.

bitbots_navigation/bitbots_localization/src/localization.cpp
Flova: I agree, but I need to keep a reference, so it is not destroyed or optimized away. But then clang format shouts at me for having a unused variable.

bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/world_model_capsule.py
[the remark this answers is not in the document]
jaagut: yes, please do so

bitbots_navigation/bitbots_localization/src/localization.cpp
jaagut: and what do the double brackets do?

Flova: The [[ ... ]] notation is known as an attribute specifier in C++. Attributes provide a standardized way to convey additional information to the compiler, which can influence optimization, code generation, and other aspects of compilation. The [[ ... ]] syntax is introduced in C++11 as a way to standardize attributes across different compilers.

jaagut (approved):

Issue #478, opened 2024-06-12
Use dynamic reconfigure in path planning


val-ba:
# Summary

PR #479, opened 2024-06-13
Add path_planning_config.yaml for dynamic reconfigure

val-ba:
# Summary

We also want to add dynamic reconfigure to path planning.
## Proposed changes

Implement https://github.com/PickNikRobotics/generate_parameter_library.
## Related issues

#478
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_navigation/bitbots_path_planning/config/path_planning_config.yaml
+++ b/bitbots_navigation/bitbots_path_planning/config/path_planning_config.yaml
@@ -1,0 +1,141 @@
+bitbots_path_planning:
+  base_footprint_frame: 
+    type: string
+    default_value: base_footprint
+    description: 'The frame of the robot base'
+    read_only: true
+  tf_buffer_duration:
+    type: double
+    default_value: 5.0
+    description: '-'
+    validation:
+      bounds<>: [0.0, 10.0]
+  rate:
+    type: double
+    default_value: 20.0
+    description: 'The rate at which the path planning is executed'
+    validation:
+      bounds<>: [0.0, 100.0]
+  
+  map:
+    planning_frame:
+      type: string
+      default_value: map
+      description: 'The frame in which the path planning is done'
+    resolution:
+      type: int
+      default_value: 20
+      description: 'Pixel per meter'
+      validation:
+        bounds<>: [0, 100]
+    size:
+      x:
+        type: double
+        default_value: 11.0
+        description: 'The size of the map in x direction'
+        validation:
+          bounds<>: [0.0, 100.0]
+      y:
+        type: double
+        default_value: 8.0
+        description: 'The size of the map in y direction'
+        validation:
+          bounds<>: [0.0, 100.0]
+    ball_update_topic:
+      type: string
+      default_value: ball_relative_filtered
+      description: 'The topic to subscribe to for ball detections'
+      read_only: true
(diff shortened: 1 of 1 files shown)


bitbots_navigation/bitbots_path_planning/bitbots_path_planning/controller.py
```diff
@@ -70,8 +52,8 @@ def step(self, path: Path) -> tuple[Twist, PointStamped]:
         # End conditions with an empty or shorter than carrot distance path need to be considered
         if len(path.poses) > 0:
             end_pose: Pose = path.poses[-1].pose
-            if len(path.poses) > self.config_carrot_distance:
-                goal_pose: Pose = path.poses[self.config_carrot_distance].pose
+            if len(path.poses) > self.node.config.carrot_distance:
```
Flova, line 55: This is missing the controller namespace

bitbots_navigation/bitbots_path_planning/bitbots_path_planning/map.py
```diff
@@ -19,19 +19,19 @@ class Map:
     def __init__(self, node: Node, buffer: tf2.Buffer) -> None:
         self.node = node
         self.buffer = buffer
-        self.resolution: int = self.node.declare_parameter("map.resolution", 20).value
-        self.size: tuple[float, float] = (
-            self.node.declare_parameter("map.size.x", 11.0).value,
```
Flova, line 24: We could use the field size config + some padding instead

Flova:
TODO integrate the comment change from #482

bitbots_navigation/bitbots_path_planning/bitbots_path_planning/controller.py
val-ba: true

bitbots_navigation/bitbots_path_planning/bitbots_path_planning/map.py
val-ba: They are still in the ros params format. How would you use these values? Additionally we would need to change it depending on which field we are playing. But It would be nice to not copy everything.

bitbots_navigation/bitbots_path_planning/config/path_planning_parameters.yaml
```diff
@@ -105,7 +105,7 @@ bitbots_path_planning:
     smoothing_tau:
       type: double
       default_value: 0.5
-      description: 'This is the time constant of the exponential smoothing filter. The higher the value, the more smoothing is applied. It denotes the time after which only 63% (1-1/e) of the original signal is left. It is update rate independent!'
+      description: 'This is the time constant of the exponential smoothing filter. It denotes the time after which only 63% (1-1/e) of the original signal is left. In other words, it denotes the time it takes for a new input to be 63% integrated into the output. It is update rate independent!'
```
Flova, line 108: This is not the text from the config. It looks similar, but you only copied part of it. Please copy the complete comment over.

Flova (changes requested):

Flova (changes requested):
See comment

Flova (approved):

Flova (approved):

PR #480, opened 2024-06-13
Handle negative observations / filter rework

val-ba:
# Summary

Checks if we should see the ball and bumps covariance if we don't.
Also removes kalman filter as we don't have too much noise currently and is simplifies the code and enables more non linear modelling of the ball behavior.

## Proposed changes

- Create a subscription on the camera info topic to get the camera matrix
- test whether it is in fov by transforming it into the camera frame and applying the camera matrix
## Related issues

#230
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -207,6 +231,50 @@ def filter_step(self) -> None:
...
                 self.publish_data(state_vec, huge_cov_mat)
 
+    def is_ball_in_fov(self, header: Header) -> bool:
+        """
+        Calculates if a ball should be currently visible
+        """
+        # Check if we got a camera info to do this stuff
+        if self.camera_info is None:
+            self.logger.info("No camera info received. Not checking if the ball is currently visible.")
+            return False
```
jaagut, line 241: I think as a fallback, we should return True. Or if this happens for a short while we wouldn't use an otherwise correct ball detection.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -207,6 +231,50 @@ def filter_step(self) -> None:
...
+            p = [
+                ball_in_camera_optical_frame.pose.position.x,
+                ball_in_camera_optical_frame.pose.position.y,
+                ball_in_camera_optical_frame.pose.position.z,
+            ]
+            k = np.reshape(self.camera_info.k, (3, 3))
+            p_pixel = np.matmul(k, p)
+            p_pixel = p_pixel * (1 / p_pixel[2])
+            # Make sure that the transformed pixel is inside the resolution and positive.
+            if 0 < p_pixel[0] <= self.camera_info.width and 0 < p_pixel[1] <= self.camera_info.height:
```
jaagut, line 274: @Flova is it the case, that YOEO could predict a bounding box with a bottom center point outside the image (e.g. because of knowledge about the ball right at the image edge)? In that case, we would dismiss an otherwise good ball.
Anyway, do we want to add a bit of tolerance here?

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
Flova: I think saying we should not be able to see the ball is the more reasonable fallback here, as it essentially reverts to the old behavior.

Flova: We only check if we should see the ball but don't see it. This is all it does. So we don't ignore any balls. It just increases the uncertainty if there is no ball but there should be one.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
Flova: I agree that we might want to add some border here, but for different reasons. As explained earlier this is not a detection by YOEO, but instead the expected ball position projected into the camera to check if we should see the ball when YOEO detects no ball. In this case the uncertainty would be increased to model the negative observation. I still think that a border makes sense. Just in the other direction. Detections near the image border might not be as reliable, so we should not treat the absence of a detection as a negative observation here. Therefore I would lower the window in image space that is used for negative observations.

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -109,6 +119,20 @@ def ball_callback(self, msg: BallArray) -> None:
...
             if position is not None:
                 self.ball = BallWrapper(position, msg.header, ball_msg.confidence.confidence)
+        # If we do not see a ball, but the KF predicts the ball to be there:
+        else:
+            # Check if we should be able to see the ball
+            if self.is_ball_in_fov(msg.header):
+                self.logger.info(
+                    "The KF prediction expects a ball, but we see none... Process noise will be increased."
+                )
+                self.kf.Q = Q_discrete_white_noise(
```
Flova, line 129: Process noise is only increased and never decreased.

Flova (changes requested):

bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
val-ba: So I should apply I threshold?

Flova (approved):

PR #486, opened 2024-06-16
Fix simulation (Kick and initial fall) 

Flova:
# Summary
Currently, we are no able to kick and fall down frequently in the simulation. This fixes that by adding a different startup animation and slightly different kick animation. 

This needs to be tested on the real robot and we might need different animations for sim and real.

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


bitbots_wolfgang/wolfgang_animations/animations/kick/kick_left.json
```diff
@@ -36,7 +36,7 @@
                 "LHipYaw": -1.14,
                 "LKnee": 56.51,
                 "RAnklePitch": 35.86,
-                "RAnkleRoll": -5.36,
+                "RAnkleRoll": 5.36,
```
val-ba, line 39: How do you know it is positive? Did the left and right ankles get mixed up? Or is that the change you made to the kick?

Flova: This is the change I made for the kick to work in sim

val-ba (approved):
I tested this in simulation, and the robot does not fall in the beginning anymore and it also performed kicks without falling.

Flova:
This was not intended to be merged without testing the changes also on the real robot.

Flova:
But I might test them later today

Issue #432, opened 2024-05-02
Use LEDs in a useful way


timonegk:
Did we see the ball? Does team communication work? What is your role? Are you walking to the ball?
Look at rules to see if it is allowed

Flova:
- Are we connected to the game controller 
- What is the strategy
- What is the role

PR #487, opened 2024-06-18
Use LEDs

ayin21:
# Summary
We have LEDs that can show different topics. But we don#t change them that much during the game.

## Proposed changes
Use LEDs to show the role and strategy of the robot as well as hardware problems.

## Related issues
#432 

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova (approved):
LGTM

bitbots_motion/bitbots_hcm/scripts/led.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.robot_controll_state_msg == msg:
+            return
+        self.robot_controll_state_msg = msg
+        # Change the led color accordingly
+        self.update_leds()
+
+    def update_leds(self):
+        """
+        Updates the leds based on the current state. The priority is as follows: robot control state > strategy role and action.
+        The robot controll state changes all leds to the same color. The strategy role and action changes the led on the left
```
texhnolyze, line 86: typo: robot control with a single `l`

bitbots_motion/bitbots_hcm/scripts/led.py
```diff
@@ -0,0 +1,123 @@
...
+        self.pub_core_1 = self.create_publisher(ColorRGBA, "/led1", 1)
+        self.pub_core_2 = self.create_publisher(ColorRGBA, "/led2", 1)
+
+        # Publishers for the imu leds
+        self.pub_imu_3 = self.create_publisher(ColorRGBA, "/led3", 1)
+        self.pub_imu_4 = self.create_publisher(ColorRGBA, "/led4", 1)
+        self.pub_imu_5 = self.create_publisher(ColorRGBA, "/led5", 1)
+
+        self.stratedy_msg: Strategy = Strategy()
+        self.robot_controll_state_msg: Optional[RobotControlState] = None
```
texhnolyze, line 37: `control` is written with one l

bitbots_motion/bitbots_hcm/scripts/led.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Maps the robot control state to a color
+        self.color_mapping_robot_controll_state = {
```
texhnolyze, line 40: `control` is written with one l

bitbots_motion/bitbots_hcm/scripts/led.py
```diff
@@ -0,0 +1,123 @@
...
+        self.update_leds()
+
+    def update_leds(self):
+        """
+        Updates the leds based on the current state. The priority is as follows: robot control state > strategy role and action.
+        The robot controll state changes all leds to the same color. The strategy role and action changes the led on the left
+        of the imu and coreboard and the stategy action changes the mittle and right led on the imu and coreboard.
+        """
+        if (
+            self.robot_controll_state_msg is not None
```
texhnolyze, line 90: Is there a way in python for `.` notation access to properties, to just return None for the accessed member?
Like typescripts `config?.value`?

Or would I need a nullable object like this:
```
config = {
  option: {
    suboption: {
      value: "test" 
    }
  }
}
```
with `if config is not None and config.option is not None and config.option.suboption is not None` before I can ever access `value`?

bitbots_msgs/msg/RobotControlState.msg
```diff
@@ -22,8 +22,6 @@ uint8 SHUTDOWN=6
 # The robot is in penalty position
 # It can not be controlled
 uint8 PENALTY=7
-# The robot is getting in or out of penalty position
-uint8 PENALTY_ANIMATION=8
```
texhnolyze, line 26: Are we not using this anymore?

texhnolyze (changes requested):
Looks great. Just a typo/naming issue for `control`

bitbots_motion/bitbots_hcm/scripts/led.py
Flova: The latter is the case and there is no standard was to handle it in a more elegant manner.

bitbots_msgs/msg/RobotControlState.msg
Flova: That's the case

texhnolyze (approved):

PR #492, opened 2024-06-20
Fix start movement in walking

timonegk:
# Summary
Fixes the start movement in walking. Before, there was a jump in the trajectory for both feet's y axis when starting the first step.

Before:
image

After:
Screenshot from 2024-06-20 15-49-13

- [x] tested in visualization
- [ ] tested on robot


timonegk:
This behavior is not nice but it makes the walking more stable... closing.

PR #501, opened 2024-06-26
Covariance based ball seen

Flova:
# Summary
Use covariance instead of timings to determine if we saw the ball recently. This cleans up the world model capsule significantly. Also simplify some minor things.

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/world_model_capsule.py
+++ b/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/world_model_capsule.py
@@ -5,6 +5,7 @@ import numpy as np
 import tf2_ros as tf2
 from bitbots_utils.utils import get_parameters_from_other_node
 from geometry_msgs.msg import (
+    Pose,
     PoseStamped,
     PoseWithCovarianceStamped,
     TransformStamped,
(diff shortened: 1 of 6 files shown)


bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/world_model_capsule.py
```diff
@@ -131,58 +122,50 @@ def get_ball_position_uv(self) -> Tuple[float, float]:
...
+        """
+        Handles incoming ball messages
+        """
+        assert msg.header.frame_id == self.map_frame, "Ball needs to be in the map frame"
+
+        # Save ball
+        self._ball = PointStamped(
+            header=Header(
+                # Set timestamps to zero to get the newest transform when this is transformed later
+                stamp=Time(clock_type=ClockType.ROS_TIME).to_msg(),
```
jaagut, line 153: There are uses where this is not transformed. And perhaps later, we want to use thie ball time again. Then it would be confusing, why the time is zero. I would prefer to set the zero time stamp only right before transforming it for your intended case (only get_ball_position_uv(), right?).

jaagut (approved):

PR #503, opened 2024-06-27
Rework node referencing and pid nodes in walking

### LINKED ISSUE #308: refactor PID controller
When https://github.com/ros-controls/control_toolbox/pull/125 is merged, it should be possible to use the same node for the PID controller and the other stuff. We should refactor our code to get rid of unnecessary nodes.

Flova:
# Summary
Pass reference of a node to the walking instead of inheriting from a node. This fixes memory management issues when passing a reference of the node to subcomponents. We also remove the two PID nodes, as they are not needed anymore.

## Related issues
Closes #308 

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


SammyRamone (approved):
lgtm

PR #507, opened 2024-06-27
Feature: publish repo status

### LINKED ISSUE #443: Publish current software version and record it in bag
# Summary

We should publish our current software version and record it in rosbags.

Recording the software versions enables us to precisely reconstruct which code ran during a game on each robot, even a long time after the game.

The published information *should* include:
- Commit hash of `bitbots_main`
- Friendly commit name of `bitbots_main`

The published information *could* include:
- Versions/commits of `lib/`
- Versions of PIP dependencies
- Versions of ROS 2 dependencies
- Versions of OS

Our live monitoring can then show a warning if the versions between running robots differ.

As we do not copy the `.git` directories to the robots, we need to preserve the needed information some other way. One possibility could be: the deploy-tool writes a simple file (maybe JSON?) to somewhere. This file is then used, to publish the (latched) topic.

It appears that ROS 2 does not provide a message-type for this purpose. We could construct a custom message-type for this or just use the String-message and dump the information in there…

@Flova @timonegk @texhnolyze What are your thoughts?

jaagut:
# Summary

This publishes the current commit hash (and friendly name) of all repositories.
This also gets recorded into rosbags.

## Proposed changes

- Deploy tool
  - Do repo-checks on `bitbots_main` and allgit-repos as defined in the VCS `workspace.repo` file in parallel
  - Print fancy warnings
  - Write `workspace_status.json` file containing all commit-hashes and friendly names
- `highlevel.launch` launches new node to read `workspace_status.json` file and publish contents as a `String` message
- rosbag-record tool records this message

## Related issues

Closes #443

## Checklist

- [X] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [X] Test on the robot


bitbots_misc/bitbots_bringup/launch/highlevel.launch
```diff
@@ -89,4 +90,14 @@
             <arg name="sim" value="$(var sim)" />
         </include>
     </group>
+
+    <!-- launch workspace status publisher -->
+    <group if="$(var workspace_status)">
+        <group unless="$(var sim)">
+            <node pkg="bitbots_utils" exec="publish_workspace_status.py" output="screen">
+                <param name="workspace_status_path" value="$(find-pkg-share bitbots_utils)/config/workspace_status.json"/>
```
Flova, line 98: If that file is always in the share of utils, you can just hard code it in Python, as the package share path is available there. It would be slightly less complex, but in the end it really doesn't matter that much.

bitbots_misc/bitbots_utils/scripts/publish_workspace_status.py
```diff
@@ -0,0 +1,53 @@
...
+        self.declare_parameter("workspace_status_path", rclpy.Parameter.Type.STRING)
+        self.declare_parameter("publish_topic", "/workspace_status")
+        self.workspace_status_path: str = self.get_parameter("workspace_status_path").value
+        self.publish_topic: str = self.get_parameter("publish_topic").value
+
+        if not os.path.exists(self.workspace_status_path):
+            self.get_logger().warning(
+                f"The workspace status file does not exist under the given path: '{self.workspace_status_path}'"
+            )
+            self.destroy_node()
```
Flova, line 25: Does destroy node terminate the code. Because there is no return after it. So it could be that it just destroys the context, but we keep executing.

bitbots_misc/bitbots_utils/scripts/publish_workspace_status.py
```diff
@@ -0,0 +1,53 @@
...
+            String, self.publish_topic, qos_profile=QoSProfile(depth=1, durability=DurabilityPolicy.TRANSIENT_LOCAL)
+        )
+
+        self.pub.publish(String(data=self.workspace_status))
+
+
+def main(args=None):
+    rclpy.init(args=args)
+    node = WorkspaceStatusPublisher()
+    executor = SingleThreadedExecutor()
```
Flova, line 41: I think you can use the default executor, which is already present and single threaded, using `rclpy.spin(node)`.

requirements/common.txt
```diff
@@ -1,6 +1,7 @@
 # This file includes common dependencies for all environments
 pip
 transforms3d==0.4.1
+PyYAML
```
Flova, line 4: I think we mainly used the requirements txt for python things that aren't in apt (or don't have an up to date version there). Consider just adding to the the package XML of Bit-Bots utils.

Flova (changes requested):
Most things are minor issues and up to personal preference, but I think pyyaml should come from rosdep/apt if possible (like it is declared in other nodes requiring it).

requirements/common.txt
Flova: Ohh it's there because of the deploy tool right? Does the deploy tool still support non Ros envs?

But we should pin the version to the apt version in this case so nothing breaks by installing the newer one.

jaagut: What are Ros envs?

Flova: Environments without Ros (rosdep, ...).

Flova (approved):

PR #513, opened 2024-06-29
ros_control: add delay before motor write after power is turned on

timonegk:
# Summary
This is necessary because the motors need some time for their startup. When ROM/RAM values are written directly after startup, they don't take effect. The start_delay parameter is added to set this delay. From testing, a value of 10ms would be sufficient but I chose a higher value to be on the safe side and because this is only rarely executed.

## Related issues
This was a problem before because custom PID values were not set after the motor power was turned off, leading to worse performance of the walking and standup.

## Checklist
- [x] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [x] Test on the robot


bitbots_lowlevel/bitbots_ros_control/include/bitbots_ros_control/wolfgang_hardware_interface.hpp
```diff
@@ -33,6 +33,8 @@ class WolfgangHardwareInterface {
   std::vector<std::vector<bitbots_ros_control::HardwareInterface *>> interfaces_;
   DynamixelServoHardwareInterface servo_interface_;
   rclcpp::Publisher<bitbots_msgs::msg::Audio>::SharedPtr speak_pub_;
+  rclcpp::Time motor_start_time_;
+  bool motor_first_write_;
```
Flova, line 37: Maybe add default value, even tho it will most likely not be relevant.

Flova (approved):

PR #519, opened 2024-07-01
pose_check.py: add ankles and zero position, reduce number of poses

timonegk:
# Summary
* add ankle check to script
* add zero-position to script (important to check horn mount)
* reduce number of poses by combining head and arms

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_lowlevel/bitbots_ros_control/scripts/pose_check.py
+++ b/bitbots_lowlevel/bitbots_ros_control/scripts/pose_check.py
@@ -379,24 +379,6 @@ def main():
             f"{Style.BRIGHT}Press enter to continue.{Style.RESET_ALL}"
         )
 
-        ########
-        # Zero #
-        ########
-
-        input(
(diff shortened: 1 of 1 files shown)


Flova:
Looks good, but I am not sure what you mean with zero position and why no offsets are applied there. The offsets are there so that all limbs are straight if the motors are at their zero position, which is e.g. easier to validate for the knees, than that slightly awkward angle they normally have at zero.

Flova:
Ping @timonegk

timonegk:
The zero position is so that you can check at all motor horns whether they are correctly mounted. In this position the screws are aligned with the motor. This makes it easier to know which motor is correct if for example the legs are different but you don't know which one is correct.

Flova:
Okay, but the offsets are for the exact same purpose, but using the overall robot geometry instead of the robot geometry as a reference, which should be easier to see.

timonegk:
I removed the zero position.

Flova (approved):

PR #527, opened 2024-07-08
Use imu to estimate rotation

val-ba:
# Summary

Right now it is rotated around the base link, so it does not work completely yet
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


val-ba:
out of date

PR #534, opened 2024-07-14
Limit diagonal speed in path planning

### LINKED ISSUE #366: Limit diagonal speed in path planning
# Summary

Currently we can go full speed forwards and sideways at the same time. I think it would be reasonable to limit that based on the maximum forward/sideward movement. I don't know if a kite or ellipse shape for the maximum speed envelope would be suited best.

Sometimes steps seem unreasonably large when we go diagonally. I observed this on the real robot at the integration test.

Not only limit the movement speed along the x and y direction (while keeping the direction). We can also interpolate the max speed based on the angle.

Lower the x and y speed. We accept slower speeds.

Flova:
# Summary
Closes #366 

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [x] Test on the robot


bitbots_navigation/bitbots_path_planning/test/__snapshots__/test_controller.ambr
```diff
@@ -1,11 +1,11 @@
 # serializer version: 1
 # name: test_default_setup
   dict({
-    'carrot_distance': 20,
+    'carrot_distance': 4,
```
texhnolyze: I think when splitting PRs you forgot to add the changed config, otherwie the limits here should not have changed.

texhnolyze (approved):

PR #544, opened 2024-07-14
Fix HCM hardware error detection

Flova:
# Summary
Due to a namespace change, the HCM hardware error detection was broken. 
Also, the motors were turned off too fast when we throw ourselves to the front in case of an overload error.

Now the hardware errors are shown correctly using the LEDs if we e.g. unplug a motor.

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [x] Test on the robot

### DIFF
--- a/bitbots_motion/bitbots_hcm/bitbots_hcm/humanoid_control_module.py
+++ b/bitbots_motion/bitbots_hcm/bitbots_hcm/humanoid_control_module.py
@@ -147,14 +147,14 @@ class HardwareControlManager:
         """Updates the diagnostic state."""
         status: DiagnosticStatus
         for status in msg.status:
-            if "//Servos/" in status.name:
+            if "/Servos/" in status.name:
                 if status.level == DiagnosticStatus.ERROR and "Overload" in status.message:
                     self.blackboard.servo_overload = True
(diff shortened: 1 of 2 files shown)


jaagut (approved):

bitbots_motion/bitbots_hcm/bitbots_hcm/humanoid_control_module.py
```diff
@@ -147,14 +147,14 @@ def diag_cb(self, msg: DiagnosticArray):
         """Updates the diagnostic state."""
         status: DiagnosticStatus
         for status in msg.status:
-            if "//Servos/" in status.name:
+            if "/Servos/" in status.name:
                 if status.level == DiagnosticStatus.ERROR and "Overload" in status.message:
```
texhnolyze, line 151: Since you already complained, that the hnadling of error states is done by string matching yourself, wouldn't it make sense to just add an enum to containing our handled possible states?

texhnolyze (approved):
I'm happy for you to merge this, but maybe add a "beginner friendly" issue for improving by not matching strings.

PR #547, opened 2024-07-15
Disable sync instructions when IMUs are on the bus 🚌

timonegk:
Our new IMU seems to have trouble with sync instructions, they result in delays of ~10ms.
This PR adds a flag to disable sync instructions, single reads/writes are then used instead.
I only implemented it for the functions we are actually using, for reading velocities, efforts, etc. it should be implemented in the future.

## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot


timonegk:
I will close this because it reduces the control loop frequency from ~480Hz to ~350Hz.

PR #554, opened 2024-07-16
ros_control: increase startup time afte power cycle to handle IMU boot

timonegk:
# Summary
Follow-up on #518, the IMU needs more time than 0.5 seconds. The variable names are also changed to reflect that it does not only influence the motors.

## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot

### DIFF
--- a/bitbots_lowlevel/bitbots_ros_control/config/wolfgang.yaml
+++ b/bitbots_lowlevel/bitbots_ros_control/config/wolfgang.yaml
@@ -1,6 +1,7 @@
 wolfgang_hardware_interface:
   ros__parameters:
     control_loop_hz: 500.0
+    start_delay: 2.0 # delay after the motor power is turned on until values are written, in seconds
 
     port_info:
       port0:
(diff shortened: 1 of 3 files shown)


bitbots_lowlevel/bitbots_ros_control/config/wolfgang.yaml
```diff
@@ -1,6 +1,7 @@
 wolfgang_hardware_interface:
   ros__parameters:
     control_loop_hz: 500.0
+    start_delay: 2.0 # delay after the motor power is turned on until values are written, in seconds
```
texhnolyze, line 4: I think it would make sense to also write if this is seconds or another unit.

texhnolyze (approved):

Flova (approved):

Flova:
A merge conflict needs to be resolved @timonegk

Issue #558, opened 2024-07-17
Vision fails to detect very close balls


confusedlama:
# Summary

Our vision does not detect balls close to the knees. We need to fix that in some way. For now we fixed it ad hoc in the ball filter, but ideally we would add more training data to the vision for this.

confusedlama:
Revert #559 if this is solved in the vision

ayin21:
Test on pi plus

texhnolyze:
This worked with our latest vision RF-DETR setup.

PR #559, opened 2024-07-17
No negative observations for close balls

confusedlama:
# Summary

Our vision sometimes doesn't detect balls if they are close to the knees. This ignores negative observations there so we don't forget the ball if we approach it. 

This is a hack for the WM and needs to be fixed in the vision.

## Proposed changes

Only perform negative observations in ball filter if they are further away than a threshold.

## Related issues

Quick fix for #558 

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -143,9 +143,22 @@ def ball_callback(self, msg: BallArray) -> None:
...
+        )
+        if estimate_relative is None:
+            return
+
+        distance_to_estimate = np.linalg.norm(numpify(estimate_relative.point))
+
         # If we did not get a ball measurement, we can check if we should have seen the ball
         # And increase the covariance if we did not see the ball
-        if not ball_measurement_updated and self.is_estimate_in_fov(msg.header):
+        # Due to vision issues with very close balls we don't do this for close balls
```
jaagut, line 160:
```suggestion
        # Due to vision issues with very close balls we don't do this for close balls  # TODO Remove with 558
```

jaagut (approved):

PR #574, opened 2024-07-19
Utilize walk stabilization

Flova:
# Summary
Stabilize trunk x position with a pitch based pd controller. This allows us to walk a lot faster while being significantly more stable.


timonegk:
Could you rename the controller from pitch to offset?

ayin21 (approved):

Flova:
> Could you rename the controller from pitch to offset? 

I thought about that, but I refactor this later and pitch it not incorrect, as it is still based on the pitch.

PR #585, opened 2024-07-19
body_behavior: use kick

timonegk:
# Summary
This uses the kick in the body behavior. Ball kick area must still be tested, kick must be tested on other robots.

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


timonegk:
This change seems to work but there is no improvement compared to the walk kick because it does not kick super far and takes some time to perform

PR #588, opened 2024-07-20
feat(behavior): descend with goalie on block/role-position

texhnolyze:
by creating a generalized cancelable `AbstractDynupAnimation` class,
where each implementing class has to define a `reset_animation` method,
which is triggered on pop

`GetWalkready` is also now using this as a base in our behavior.

In the future this class should be merge with the similar
`PlayAnimationDynup` class in HCM.

## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
Is this still a draft or ready for review? @texhnolyze

jaagut:
Ping @texhnolyze

texhnolyze:
This is still up to date, but requiress some more testing and has a few todos.
I will handle it.

jaagut:
Currently outdated. @texhnolyze do you think, this is still useful for a non-Wolfgang future?

PR #589, opened 2024-07-20
team_communication: automatically detect target ip address

### LINKED ISSUE #578: Dynamically decide team communication IP
# Summary
Currently, we configure the team communication IP in a config file. This way we have to change it between competitions, lab, etc. to use the broadcast IP of the network in most cases.

A suggestion would be to just dynamically use the broadcast IP of the connected wifi interface automatically, but also keep the option to manually configure it.
Additionally from ros 1 times we also have a local mode, which was used when simulating multiple robots on a single machine, which is not possible anymore, and can be removed.

timonegk:
# Summary
Automatically detect the target address for team communication from the wifi interface using `netifaces`.
In case of any problem, the default address configured is used.

## Related issues
Closes #578.

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/communication.py
+++ b/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/communication.py
@@ -1,5 +1,5 @@
 import socket
-from ipaddress import IPv4Address
+from ipaddress import AddressValueError, IPv4Address
 from typing import List, Optional
 
 from rclpy.node import Node
@@ -11,7 +11,19 @@ class SocketCommunication:
 
         self.buffer_size: int = 1024
         self.socket: Optional[socket.socket] = None
-        self.target_ip: IPv4Address = IPv4Address(node.get_parameter("target_ip").value)
+        if node.get_parameter("detect_target_ip").value:
+            try:
+                # automatically detect from subnet
+                import netifaces
+
+                self.target_ip: IPv4Address = IPv4Address(
+                    netifaces.ifaddresses(node.get_parameter("wifi_interface"))[netifaces.AF_INET][0]["broadcast"]
+                )
+            except (ImportError, ValueError, KeyError, AddressValueError):
+                self.logger.warn("Could not detect broadcast address, falling back to configured address")
+                self.target_ip = None
+        if self.target_ip is None:
+            self.target_ip: IPv4Address = IPv4Address(node.get_parameter("target_ip").value)
 
         if self.target_ip.is_loopback:
             # local mode on loopback device, bind to port depending on bot id and team id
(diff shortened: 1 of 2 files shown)


bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/communication.py
```diff
@@ -11,7 +11,19 @@ def __init__(self, node: Node, logger, team_id, robot_id):
...
+                self.target_ip: IPv4Address = IPv4Address(
+                    netifaces.ifaddresses(node.get_parameter("wifi_interface"))[netifaces.AF_INET][0]["broadcast"]
+                )
+            except (ImportError, ValueError, KeyError, AddressValueError):
+                self.logger.warn("Could not detect broadcast address, falling back to configured address")
+                self.target_ip = None
+        if self.target_ip is None:
+            self.target_ip: IPv4Address = IPv4Address(node.get_parameter("target_ip").value)
 
         if self.target_ip.is_loopback:
```
texhnolyze, line 28: Since you removed `local_target_ports` this whole block should also be removed

texhnolyze (changes requested):

bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/communication.py
```diff
@@ -11,7 +11,19 @@ def __init__(self, node: Node, logger, team_id, robot_id):
...
+                # automatically detect from subnet
+                import netifaces
+
+                self.target_ip: IPv4Address = IPv4Address(
+                    netifaces.ifaddresses(node.get_parameter("wifi_interface"))[netifaces.AF_INET][0]["broadcast"]
+                )
+            except (ImportError, ValueError, KeyError, AddressValueError):
+                self.logger.warn("Could not detect broadcast address, falling back to configured address")
+                self.target_ip = None
+        if self.target_ip is None:
```
jaagut, line 25: `self.target_ip` may not exist at this point

bitbots_team_communication/bitbots_team_communication/config/team_communication_config.yaml
```diff
@@ -1,21 +1,15 @@
 team_comm:
   ros__parameters:
-    # UDP broadcast address is the highest IP in the subnet e.g. 172.20.255.255
-    # Sets local mode if set to loopback (127.0.0.1)
+    # Automatically detect UDP broadcast address
+    detect_target_ip: true
+    wifi_interface: "wlp3s0"
+    # Fallback, only used if detect_target_ip is false. This should be the highest IP in the subnet e.g. 172.20.255.255
```
jaagut, line 6: `only used if detect_target_ip is false` this is false or irritating.

Its used:
- _if detect_target_ip is false_
- _if detect_target_ip is true, but auto-detection failed_

jaagut (changes requested):

Flova:
What's the state with this PR? @timonegk

timonegk:
It is still relevant but the reviews must be addressed and it has to be tested

PR #595, opened 2024-08-11
Let CI fail earlier if our server is down

jaagut:
# Summary

When our server is down, the wget commands in our CI job retry 20 times with a long timeout. This reduces the timeout to 15 seconds and the retries to 2. With that, our CI should fail much earlier. Also, this does not really harm the user running make install.

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,8 @@ pull-files:
 	wget \
 		--no-verbose \
 		--show-progress \
+		--timeout=15 \
+		--tries=2 \
 		--recursive \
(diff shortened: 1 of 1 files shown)


Flova:
Should it time out currently?

jaagut:
Yes

Flova (approved):
Merge if CI passed

PR #596, opened 2024-09-25
Fix walking python bindings

Flova:
The migration of the walking to the `generate_parameter_library` required a significant rework of the way parameters are set when the Python walk wrapper is used. Coincidentally the new approach simplifies the logic a bit, which is nice I guess.

I also added a few longer comments explaining the reasoning behind some design choices.


Flova:
@coderabbitai review

jaagut (approved):

PR #609, opened 2024-10-31
Update install docs

Flova:
# Summary
Fix ~ folder issue, links and ssh setup


jaagut (commented):
As code rabbit suggested, we should consider using more relative links.

Also, after the install ros step with concatenating the bashrc: one has to either
- source bashrc manually,
- close the session, reopen and the next step still assumes, we are in the ws/src/bitbots_main directory which we are not
- or call bash again

I think, sourcing is the cleanest option. We should add this to the install ros doc

Flova:
> As code rabbit suggested, we should consider using more relative links.

I agree, but I think this is out of scope for this PR.

jaagut (approved):

PR #620, opened 2024-11-11
Sim footsensor phase reset and PIDs

Flova:
# Summary
- Use foot pressure sensors in sim
- Use PID controllers in sim

### DIFF
--- a/bitbots_motion/bitbots_quintic_walk/config/walking_wolfgang_simulator.yaml
+++ b/bitbots_motion/bitbots_quintic_walk/config/walking_wolfgang_simulator.yaml
@@ -75,23 +75,23 @@ walking:
       phase_reset:
         min_phase: 0.90
         foot_pressure:
-          active: False
+          active: True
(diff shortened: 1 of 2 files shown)


bitbots_motion/bitbots_quintic_walk/config/walking_wolfgang_simulator.yaml
```diff
@@ -75,23 +75,23 @@ walking:
...
         pitch:
-          p: 0.0
+          p: 0.0035
           i: 0.0
-          d: 0.0
+          d: 0.004
           i_clamp_min: 0.0
           i_clamp_max: 0.0
-          antiwindup: False
+          antiwindup: false
```
jaagut, line 94: Why did you change this? Seems like other cases are written as `False`.

Flova: I think I just copied it from the real robots config file.

Issue #624, opened 2024-11-20
rqt animation gui value changes not saved


lvaddi:
# Summary
Changes in value boxes are not saved. For example when changing a value and switching frames, going back to the changed frame shows the old value again.

## Expected behavior
Changes in value boxes should be saved when switching frames or saving the animation.

## Steps to Reproduce

1. Start rqt animation gui
2. Create/Open animation
3. Create a second frame
4. Change a value in one of the boxes
5. Switch frames
6. Go back to the changed frame
7. Value has not been changed

## Possible Solution
Change something with the QtDoublespinboxes that are used for the boxes.

Flova:
You need to press record to preserve values. This is by design, but I understand that it can be confusing.

PR #625, opened 2024-11-20
Fix/animation gui

confusedlama:
# Summary
Discarding the unrecorded data is expected behavior when changing frames. I have added a warning that reminds the user to record changes that would be discarded.

## Related issues
https://github.com/bit-bots/bitbots_main/issues/624

## Checklist
- [ ] run simulation and motion in background
- [ ] start rqt
- [ ] change frames without changing motor goals (no warning should be displayed)
- [ ] change frames with changed motor goals (warning should be displayed)
- [ ] change frames with newly recorded motor goals (warning should not be displayed)


Flova (changes requested):
Seems good.

I would have added a method to the recorder that checks if the values have changed for a given pair of frame and working values. But you approach is also totally valid. 

I requested changes because of the failing formatting checks and the present debug prints.

Flova (approved):
Lgtm, but the ci seems to be broken. I would merge once it is fixed.

PR #628, opened 2024-11-23
Jazzy migration

Flova:


JanNiklasFeld:
TODO: Fix cppformat in Jazzy

Flova:
The CI seems to fail due to floating point errors in the team_communication.

Flova:
TODO after merge: Revert #634

jaagut:
At this point, my devcontainer (built 6 days ago) only has a small warning on `bitbots_pybullet_sim`:
```
--- stderr: bitbots_pybullet_sim
CMake Warning:
  Manually-specified variables were not used by the project:

    CATKIN_INSTALL_INTO_PREFIX_ROOT
    CATKIN_SYMLINK_INSTALL
```

Now, I noticed that my bitbots_main was not clean. Therefore, I recloned and rebuilt the devcontainer. Finally, I can see the same compile failures and warnings from the CI!

jaagut:
Since the latest apt-release, the CI and local builds fail because `realtime_tools` (transitive dependency) deprecated its `.h` header files (https://github.com/ros-controls/realtime_tools/pull/206), but the `control_toolbox`'s fix (https://github.com/ros-controls/control_toolbox/pull/247) has not yet been released. So, we have to wait for this or turn off warnings being treated as errors in the meantime. This effects `bitbots_dynup`, `bitbots_dynamic_kick`, and `bitbots_quintic_walk`.

Similar issues appear with other components, but not as errors. We have to wait for https://github.com/moveit/moveit2/pull/3197 and then update:
- `bitbots_moveit_bindings`
- `bitbots_ros_control`
- `bitbots_heaad_mover`
- `bitbots_dynup`
- `bitbots_dynamic_kick`
- `bitbots_quintic_walk`

Once this is done, all issues are resolved (except the minor one listed in my comment above)!

jaagut:
I have opened an issue #642 to turn warnings as errors back on.

jaagut:
Now, only this warning remains at compilation. Nothing has yet been executed. We should test thoroughly in sim and on robot.

> At this point, my devcontainer (built 6 days ago) only has a small warning on `bitbots_pybullet_sim`:
> 
> ```
> --- stderr: bitbots_pybullet_sim
> CMake Warning:
>   Manually-specified variables were not used by the project:
> 
>     CATKIN_INSTALL_INTO_PREFIX_ROOT
>     CATKIN_SYMLINK_INSTALL
> ```
> 
> Now, I noticed that my bitbots_main was not clean. Therefore, I recloned and rebuilt the devcontainer. Finally, I can see the same compile failures and warnings from the CI!

Flova:
> EDIT: I have migrated everything to `.hpp` (where available) in our @bit-bots repositories. And I have turned off "warnings as errors" temporarily in [65d0099](https://github.com/bit-bots/bitbots_main/commit/65d0099c4944a88e53ca3a76cb857602c2ee21c9)
> We should revert this commit, once [ros-controls/control_toolbox#247](https://github.com/ros-controls/control_toolbox/pull/247) is released in apt

We can also just build this from source for now by adding it to lib

Flova:
> Now, only this warning remains at compilation

This has been addressed in #643

jaagut:
> We can also just build this from source for now by adding it to lib

I know, but that's more effort and we get unclean lib/ but I guess, that would be resolved with a make fresh libs...

Btw, do you know, why we had `imu_tools` in our lib/? I guess it was just not fully available in the past.

Flova:
> Btw, do you know, why we had `imu_tools` in our lib/? I guess it was just not fully available in the past

I think it was not available in the package sources when we started out with ros2 or something like this.

Flova:
TODO: revert #646 on this branch if it is merged into main and this branch is updated. These are iron specific fixes.

Flova:
The second merge was weird but needed for some reason

jaagut:
I have fixed some documentation issues, more deprecation warnings and removed control_toolbox from lib/ again, as the APT version has been updated.

Unfortunatly, my change regarding the colorful colcon packages in 4640f9ec5979406665013ae06a965f794c9c2ad9 does not seem to work... Plain pip is not available anymore, and including it in the requirements with --break-system-deps does not seem to help either. Also tried pipx

Some mypy typing issues cause tests to fail.

jaagut:
Uff, why was this pyi stuff in cmakelists necessary?

Flova:
Protobuf generates Python code. That Python code makes mypy unhappy. But we can not exclude it, because ros2 fucked up it's mypy configuration by adding another layer on top of it. So we need to generate stubs for the protobuf messages, which is better either way, as you also get better ide support with it. These are the .pyi files. And because ament_mypy is broken in another way it does not recognize stub files. This was fixed last week in rolling, but it is not available yet. We therefore need to manually discover all necessary files in our cmake and pass them to ament_mypy

bitbots_misc/bitbots_tts/bitbots_tts/tts.py
Flova, line 1: Could we separate this into a different PR? This is a new feature and needs to be tested differently performance wise. Also things like catching might need to be implemented.

bitbots_motion/bitbots_quintic_walk/config/walking_wolfgang_simulator.yaml
```diff
@@ -75,13 +75,13 @@ walking:
       phase_reset:
         min_phase: 0.90
         foot_pressure:
```
Flova, line 77: Do we want to change this in the normal simulation env from now on. We might need to tune it for the IMU phase reset to work properly. This affects not just the hlvs setup

texhnolyze (approved):

PR #629, opened 2024-11-27
Demo behavior

Flova:
# Summary
Adds a simplified behavior that works without a field and localization. Mainly fokussed on external demos e.g. at ITCS or University events.

## Proposed changes
- Add simplified dsd file
- Fix ball close target, so we always approach the ball at the closest angle (no need to aim for the goal)
- Add a launch file

## Related issues
#

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
We tested this, but the test was rushed and Donna had hardware issues. So I would recommend a more thorough test in the near future.

Flova:
Today we have done some proper testing and it LGTM

bitbots_navigation/bitbots_path_planning/bitbots_path_planning/planner.py
```diff
@@ -94,3 +100,31 @@ def get_path(self) -> Path | None:
...
+
+        self.path = Path(
+            header=Header(frame_id=self.map.get_frame(), stamp=self.node.get_clock().now().to_msg()),
+            poses=[pose, self.goal],
+        )
+
+        return self.path
+
+
+def planner_factory(node: Node) -> type:
```
jaagut, line 126:
```suggestion
def planner_factory(node: Node) -> Planner:
```
I think, the return type should be Planner

jaagut (approved):

bitbots_navigation/bitbots_path_planning/bitbots_path_planning/planner.py
Flova: No we return a type, that is instantiated by the user :D

Issue #598, opened 2024-10-09
Try out Cubic Splines in Quintic walk


jgueldenstein:
# Summary
We discussed during the software planning meeting that we would like to try using Cubic splines in quintic walk.
While this would require a rename and increase jerk of the trajectory, it could help to walk more efficiently with less accelerations required.

val-ba:
#632 is the pr to this issue

PR #632, opened 2024-12-06
Use cubic splines in walking

val-ba:
# Summary

Switch from quintic to cubic. We do not set the acceleration anymore and just need 4 coefficients to set the spline.
Also a lot of the time the acceleration is not even set
## Proposed changes

Cubic splines look like this for walking:
Screenshot from 2024-12-13 11-30-24
Screenshot from 2024-12-13 11-31-22

Fairly similar to the quintic splines:
Screenshot from 2024-12-13 11-52-22
Screenshot from 2024-12-13 11-52-59

## Related issues

 #598
## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


val-ba:
image
This is cubic splines right ankle and knee, we see some jumps when the robot starts to walk in the motor goals. This is the same for quintic splines aswell:
image

val-ba:
As this does not seem to either improve or worsen performance, we leave it as is and do not merge it. But for now we keep the branch.

jaagut:
We can also just restore the branch from this PR. I am deleting it now to reduce clutter.

PR #639, opened 2024-12-20
Add runtime and static type checks

Flova:
# Summary

Many of our typehints are wrong. This adds `beartype` based type checks during runtime and static `mypy` checks to prevent this.

## Proposed changes
- Add beartype
- Fix errors
- Fit unrelated crash in animation ui that made testing hard.
- Add mypy base tests

This adds some risk, as it might lead to a less frequently visited code part with a minor type error crashing our codebase during a game. We could downgrade the errors to warnings, but then they might not be noticed/ignored. We could also deactivate them during real games, but we do not have a mechanic to check for this yet.

We need to test this on the real robot before merging it!

## Checklist

- [x] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


Flova:
Ping :)

.vscode/settings.json
```diff
@@ -221,9 +221,11 @@
     "ros.distro": "iron",
     "search.useIgnoreFiles": false,
     "python.autoComplete.extraPaths": [
+        "/opt/openrobots/lib/python3.10/site-packages",
```
jaagut, line 224: Should this be here?

bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/__init__.py
```diff
@@ -9,6 +10,7 @@
 class AbstractBlackboardCapsule:
     """Abstract class for blackboard capsules."""
 
-    def __init__(self, node: Node, blackboard: Optional["BodyBlackboard"] = None):
+    @nobeartype
```
jaagut, line 13: why is nobeartype needed?

bitbots_behavior/bitbots_blackboard/package.xml
```diff
@@ -37,6 +37,10 @@
     <exec_depend>std_srvs</exec_depend>
     <exec_depend>tf2_geometry_msgs</exec_depend>
     <exec_depend>tf2</exec_depend>
+    <test_depend>ament_mypy</test_depend>
+    <test_depend>ament_cmake_mypy</test_depend>
```
jaagut, line 41: why did you depend on both? In other cases, you only use one of them

bitbots_simulation/bitbots_webots_sim/bitbots_webots_sim/webots_robot_controller.py
```diff
@@ -35,9 +36,9 @@ def __init__(
         :param external_controller: Whether an external controller is used, necessary for RobotSupervisorController
```
jaagut, line 36: The param comments are very outdated

jaagut (approved):
Uff

bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/__init__.py
Flova: This has something to do with circular imports. The blackboard is imported for static type checking, but importing it for dynamic type checking leads to a circular dependency during runtime. Afaik bear type has no mechanism to handle this.

bitbots_behavior/bitbots_blackboard/package.xml
Flova: Maybe because this is a cmake python package and the other ones are setup.py. I don't remember.

.vscode/settings.json
Flova: Not really, but VS code always adds it and I dont want to make a special rule.

PR #650, opened 2025-01-07
Account for measurement delay in localization

Flova:
# Summary
Until now we rated the current state of the localization particles always with a delayed measurement. The particles are moved using the odometry "instantly" and are therefore not where the localization expects them to be at a different place compared to when the measurement was taken. Therefore, particles that lag behind were preferred, as they better matched the observations, which were assumed to be also instantly. 

The solution is to store the observation timestamp and move all particles temporary back by the amount we have walked since the measurement, as determined by the odometry before rating them. This solution works also for multiple asynchronous measurement sources. 

I hope that this enables more aggressive path planning parameters with less oscillations, as the localization state is not delayed as much.

I also did some cleanup, like simplifying the motion model math significantly and removing the field boundary input (we never used it).

Here is a plot of the robots center of mass and the position estimated by the localization. It needs to be noted that the robot walks very fast and the vision had an artificial pipeline delay of 1 second to test the effectiveness of the approach:

Screenshot from 2025-01-07 19-18-14

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/ball_kick_area.py
+++ b/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/ball_kick_area.py
@@ -25,6 +25,7 @@ class BallKickArea(AbstractDecisionElement):
         ball_position = self.blackboard.world_model.get_ball_position_uv()
 
         self.publish_debug_data("ball_position", {"u": ball_position[0], "v": ball_position[1]})
+        self.publish_debug_data("smoothing_close", f"{self.no_near_decisions} ({self.no_near_decisions / self.smoothing * 10}%)")
 
         # Check if the ball is in the enter area
         if 0 <= ball_position[0] <= self.kick_x_enter and 0 <= abs(ball_position[1]) <= self.kick_y_enter:
--- a/bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml
+++ b/bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml
@@ -146,7 +146,7 @@ body_behavior:
     dribble_ball_distance_threshold: 0.5
     dribble_kick_angle: 0.6
 
-    kick_decision_smoothing: 5.0
+    kick_decision_smoothing: 1.0
 
     ##################
     # costmap params #
--- a/bitbots_navigation/bitbots_localization/include/bitbots_localization/ObservationModel.hpp
+++ b/bitbots_navigation/bitbots_localization/include/bitbots_localization/ObservationModel.hpp
@@ -6,17 +6,15 @@
 #define BITBOTS_LOCALIZATION_OBSERVATIONMODEL_H
 
 #include <particle_filter/ParticleFilter.h>
+#include <tf2_ros/buffer.h>
 
 #include <bitbots_localization/RobotState.hpp>
 #include <bitbots_localization/map.hpp>
 #include <bitbots_localization/tools.hpp>
 #include <sensor_msgs/msg/point_cloud2.hpp>
 #include <sensor_msgs/point_cloud2_iterator.hpp>
-#include <soccer_vision_3d_msgs/msg/field_boundary.hpp>
 #include <soccer_vision_3d_msgs/msg/goalpost.hpp>
 #include <soccer_vision_3d_msgs/msg/goalpost_array.hpp>
-#include <soccer_vision_3d_msgs/msg/marking_array.hpp>
-#include <soccer_vision_3d_msgs/msg/marking_intersection.hpp>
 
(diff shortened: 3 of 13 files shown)


bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml
```diff
@@ -146,7 +146,7 @@ body_behavior:
     dribble_ball_distance_threshold: 0.5
     dribble_kick_angle: 0.6
 
-    kick_decision_smoothing: 5.0
+    kick_decision_smoothing: 1.0
```
Flova, line 149: I noticed that this is not really needed anymore btw/

bitbots_navigation/bitbots_localization/include/bitbots_localization/localization.hpp
```diff
@@ -47,7 +47,6 @@
 #include <rclcpp/rclcpp.hpp>
 #include <sensor_msgs/msg/camera_info.hpp>
 #include <sensor_msgs/msg/point_cloud2.hpp>
-#include <soccer_vision_3d_msgs/msg/field_boundary.hpp>
 #include <soccer_vision_3d_msgs/msg/goalpost_array.hpp>
```
val-ba, line 50: same here, are goalposts even needed?

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -27,9 +27,6 @@ Localization::Localization(std::shared_ptr<rclcpp::Node> node)
   goal_subscriber_ = node->create_subscription<sv3dm::msg::GoalpostArray>(
```
val-ba, line 27: same here, goal posts needed?

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -203,11 +190,9 @@ void Localization::LinePointcloudCallback(const sm::msg::PointCloud2 &msg) { lin
 
 void Localization::GoalPostsCallback(const sv3dm::msg::GoalpostArray &msg) { goal_posts_relative_ = msg; }
```
val-ba, line 191: same here

bitbots_navigation/bitbots_localization/src/localization.cpp
```diff
@@ -355,15 +353,9 @@ void Localization::getMotion() {
         linear_movement_.y = 0;
         RCLCPP_WARN(node_->get_logger(), "Robot moved an unreasonable amount, dropping motion.");
       }
-
-      // Check if robot moved
```
val-ba, line 359: is robot_moved not checked anymore? Is this not important?

val-ba:
I tested this in sim on my machine and it worked.

bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/ball_kick_area.py
```diff
@@ -25,6 +25,9 @@ def perform(self, reevaluate=False):
         ball_position = self.blackboard.world_model.get_ball_position_uv()
 
         self.publish_debug_data("ball_position", {"u": ball_position[0], "v": ball_position[1]})
+        self.publish_debug_data(
+            "smoothing_close", f"{self.no_near_decisions} ({self.no_near_decisions / self.smoothing * 10}%)"
```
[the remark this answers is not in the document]
Flova: It smooths the ball close decision.

bitbots_navigation/bitbots_localization/include/bitbots_localization/ObservationModel.hpp
```diff
@@ -6,17 +6,15 @@
...
 #include <particle_filter/ParticleFilter.h>
+#include <tf2_ros/buffer.h>
 
 #include <bitbots_localization/RobotState.hpp>
 #include <bitbots_localization/map.hpp>
 #include <bitbots_localization/tools.hpp>
 #include <sensor_msgs/msg/point_cloud2.hpp>
 #include <sensor_msgs/point_cloud2_iterator.hpp>
-#include <soccer_vision_3d_msgs/msg/field_boundary.hpp>
 #include <soccer_vision_3d_msgs/msg/goalpost.hpp>
```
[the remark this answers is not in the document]
Flova: The field boundary performs quite poorly in the localization and we might include the goalpost with a higher chance in the near future (although not planned).

bitbots_navigation/bitbots_localization/src/localization.cpp
Flova: It was not used.

val-ba (approved):

PR #666, opened 2025-03-12
Revert "Remove unnecessary dependency maturin"

JanNiklasFeld:
Reverts bit-bots/bitbots_main#665


jaagut:
This was a misunderstanding. Another PR should be reverted.

PR #683, opened 2025-03-27
Change sim fov

### LINKED ISSUE #497: Camera FOV in simulation differs from real world.
# Summary
The FOV defined in the proto file: https://github.com/bit-bots/bitbots_main/blob/22622a780bca744a302a4546160e4e9ac9f91733/bitbots_wolfgang/wolfgang_webots_sim/protos/robots/Wolfgang/Wolfgang.proto#L74 does not match the real robot's FOV.

If we want the FOV to match the real robot we should set it to 1.517 instead.

val-ba:
# Summary

Should fix #497
This is based on the jazzy dev branch though.
## Proposed changes

Change fov in wolfgang.proto from 1.04 to 1.517
## Checklist

- [x] Run `colcon build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


Flova:
I think you might have used the wrong base branch. There are a lot of changes here

PR #686, opened 2025-04-14
Replace mimic3 with piper

val-ba:
# Summary

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_misc/bitbots_tts/bitbots_tts/tts.py
+++ b/bitbots_misc/bitbots_tts/bitbots_tts/tts.py
@@ -19,9 +19,9 @@ from rclpy.publisher import Publisher
 from bitbots_msgs.msg import Audio
 
 # Load the Piper voice
-bb_tts_dir = Path(__file__).parent + "/model"  # TODO: check how to get nice relative paths
-model_path = bb_tts_dir + "/en_US-lessac-medium.onnx"
-config_path = bb_tts_dir + "/en_US-lessac-medium.onnx.json"
+bb_tts_dir = Path(__file__).parent.parent / "model"  # TODO: check how to get nice relative paths
+model_path = bb_tts_dir / "en_US-lessac-medium.onnx"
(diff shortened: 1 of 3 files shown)


Flova (changes requested):
A few things:
- How is the model downloaded?
- I think we should implement caching so the tts is not calculated every time (we had this in mimic3)
- With mimic we had this seperate server that was started by systemd (configured in Ansible) that way we could e.g. announce the IP etc. at startup

bitbots_misc/bitbots_tts/bitbots_tts/tts.py
```diff
@@ -27,10 +34,35 @@ def speak(text: str, publisher: Publisher, priority: int = 20, speaking_active:
...
+    This is also used for speaking the ip adress during startup."""
+    synthesize_args = {
+        "length_scale": 1.0,  # Phoneme length, if lower -> faster
+        "noise_scale": 0.667,  # Generator noise, if lower -> more robotic
+        "noise_w": 0.8,  # Phoneme width noise, if lower -> more robotic
+        "sentence_silence": 0.1,  # seconds of silence after each sentence
+    }
+    with io.BytesIO() as buffer:
+        with wave.open(buffer, "wb") as wav_file:
+            voice.synthesize(text, wav_file, **synthesize_args)
```
texhnolyze, line 47: This would mean that we will generate a new `wav` file each time?
`mimic` was implicitly caching the already generated voices, which was also the reason for the web server if I remember correctly.

Have you tried how long the wav generation takes?

texhnolyze (changes requested):

bitbots_misc/bitbots_tts/bitbots_tts/tts.py
val-ba: It is real time on my laptop, but I did not try it on the robot yet.

PR #717, opened 2025-08-29
Workspace repo

Flova:
# Summary
We wanted to try a setup where the bitbots_main repo is the colcon workspace itself. This would make a few workspace related settings easier, fix the vscode ROS extention, clean up the repo top level and allow us to have all commands at the same level (managed now via `just` instead of make, as it e.g. supports positional args).

## Proposed changes
- Move code to `src/`
- Adapt paths in docker, build, setup, ...
- Migrate from `make` to `just` and add additional targets like `just build` or `just clean` 

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


.devcontainer/devcontainer.json
```diff
@@ -18,12 +18,30 @@
         "dev.containers.copyGitConfig": false,
         "dev.containers.gitCredentialHelperConfigLocation": "none"
       },
-      "extensions": ["ms-iot.vscode-ros"]
+      "extensions": [
```
texhnolyze, line 21: Aren't we managing extensions by the `.vscode` settings already.
I don't have anything about predifinition in the devcontainer, but am afraid that this will not be kept consistent with the other extension definition.

src/bitbots_misc/bitbots_docs/docs/manual/tutorials/install_software_ros2.rst
```diff
@@ -110,7 +100,7 @@ In case you are not using the bash shell, replace ``~/.bashrc`` and ``bash`` wit
   export ROS_AUTOMATIC_DISCOVERY_RANGE=LOCALHOST
 
   # Set the default colcon workspace
-  export COLCON_WS="\$HOME/colcon_ws"
+  export COLCON_WS="\$HOME/git/bitbots/bitbots_main"
```
texhnolyze, line 103: This is inconsistent with `COLCON_WS=$HOME/bitbots_main` in `setup.sh`.
And you would also need to change the online command accordingly if you now expect a different directory structure.

texhnolyze (changes requested):
It looks fine.
But I don't quite understand the intention behind making the repository itself the colcon workspace.
I thought the split is quite useful for the following reasons:
- Code is independent of build/generated files, which makes searching easier and doesn't "dirty" the repository, potentially making it more cluttered
- The split allows to easily add additional colcon packages to be built, by just adding them to the `COLCON_WS/src` without dirtying the repo (e.g. I used this when trying out ros-rust)
- The directory structure is more independent as you can just expect a `colcon_ws` workspace in scripts, but the actual code can be anywhere as it's just a symlink

Flova:
Thanks for having a look :)

> - Code is independent of build/generated files, which makes searching easier and doesn't "dirty" the repository, potentially making it more cluttered

I agree that having the `build` and `install` folders  can lead to a bit of clutter, but when navigating the workspace during use the folder structure is nearly identical to a normal workspace setup and due to these folders being git ignored, the upstream repo is tidy, maybe even more so since all code now lives in a source folder, which would otherwise result in an additional nesting. 
Also afaik many build systems like e.g. cargo store their artifacts inside an ignored folder of the repo by default, so it should be familiar for many people. The code lives in the source folder which can still be searched. The folder structure stays the same. Essentially only the `.git` folder moves upwards.

> - The split allows to easily add additional colcon packages to be built, by just adding them to the `COLCON_WS/src` without dirtying the repo (e.g. I used this when trying out ros-rust)

We already do this using the lib folder inside the repo. All third party packages are already placed there. And for non Bit-Bots related things, like e.g. a fix for ros2 itself I always create a second workspace to isolate things. There are also many cases where having the colcon workspace as a repo is a bad idea imo.. E.g. at tams there are many repos for the different robots, sensors etc. and if you do an experiment you clone them inside your colcon workspace. But with the mono repo I think it is okay to couple the Bit-Bots workspace together with our code.

> - The directory structure is more independent as you can just expect a `colcon_ws` workspace in scripts, but the actual code can be anywhere as it's just a symlink

What would be benefits of symlinked code? I know some people do this, but I do not see that many practical advantages. I would argue that it is a good thing that the directory structure is standardized by the repo, this way the scrips etc can expect a certain configuration without relying too much on global often not directly version controlled configuration.  My ideal setup would be one where you clone the repo, build it and everything works, creating minimal friction. 

I have often seen especially new people struggle with where which command should run (make commands vs. coconut) Aliases helped a bit there, but they also implicitly change directories etc.. My intention was to unify this by having one top-level repo where everything can be performed.

Also the vscode ros extension now requires that the whole workspace is open in vs code, which requires that the preconfigured settings live in the colon ws, which is unlucky as we have them inside the repo. Yeah we could symlink, but that also increases the complexity. 

Also this reduces additional setup like creating a workspace with the correct folder structure etc.. this is mostly automated right now, but it would be neat if we wouldn't need that. 

Non Ros/package related scripts can live in their own directory that is not part of the colon environment that way reducing potential friction points. 

This allows us to have a command runner that executes all relevant commands for typical usage cases.

.devcontainer/devcontainer.json
Flova: We do have a file with recommended extensions, and I agree with you, but things like the ros extension are a must have in the dev container imo. so including them in the build makes sense. The old ros extension is no longer supported, so I included the new one as well as a few essentials from the recommended ones. Ideally there would be a way of cross referencing the files, but I don't think this is the case :(

PR #730, opened 2025-12-07
Use Pixi

### LINKED ISSUE #726: Migrate from pip to uv
# Summary

Migrate from dependency management via `pip --break-system-packages` and `requirements/` files to `uv`.

`uv` is faster than pip and using a venv, we don't risk conflicts with other parts of the OS.

- Migrate over to `uv`
- Change `Makefile`/``just accordingly
- Change `setup.sh` to activate the venv with source-all `sa`
    - Idea: Change the terminal prompt, once sourced. I don't want it to always show `(.venv)` or something and we could use it as an indicator for having sourced already. Maybe the prompt could look like:
    ```shell
    🤖 <user>@<host> ׀
    ```

Flova:
### TODO

- [x] Fix events executor linking issue with chrono (see https://github.com/ros2/rclpy/pull/1563, which is already vendored by us)
- [x] Adapt docs
- [x] Test setup script

Closes #726 
Closes https://github.com/bit-bots/bitbots_main/pull/717

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      PSModulePath: ""  # Otherwise colcon might think that PowerShell is being used
 
     steps:
       - name: Cancel outdated jobs
@@ -36,9 +38,7 @@ jobs:
           cache: true
 
       - name: Build
-        run: |
-          unset PSModulePath  # Otherwise colcon might think that PowerShell is being used
-          pixi run build
+        run: pixi run build
 
       - name: Test
         run: pixi run test --parallel-workers 1


jaagut:
Closes #726
Closes #717

mikaso-stud:
Tested on Mac OS Version 26.1 (25B78)

Error:   × failed to solve requirements of environment 'default' for platform 'osx-
  │ arm64'
  ├─▶   × failed to solve the environment
  │   
  ╰─▶ Cannot solve the request because of: No candidates were found for
      libnuma >=2.0.18,<3.

pixi.toml
```diff
@@ -0,0 +1,244 @@
...
+# Patches
+pillow = "<12" # this constraint is needed as yoeo specifies the 11 major version and otherwise uv fails to solve given the constrains from the conda ecosystem.
+setuptools = "==68.1.2"  # Pinning setuptools to avoid issues with colcon build
+git = ">=2.52.0,<3"  # Otherwise the system git might be used causing ABI issues
+curl = ">=8.17.0,<9"
+make = ">=4.4.1,<5"  # To avoid issues in the CI
+
+[feature.ros.target.linux-64.dependencies]
+# For now Webots is only packaged for x86-64 Linux
+webots = ">=2022b,<2023a"
```
texhnolyze: Is it possible to define a system dependency in pixi?
I would much prefer a global installation of `webots` to 20+ local ones.
See: https://github.com/bit-bots/ansible/issues/4

PR #734, opened 2026-01-18
Fix/docs

jaagut:
# Summary

This updates several sections of our public documentation and simplifies the installation instructions a bit.

## Checklist

- [ ] Run `colcon build`
- [x] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


scripts/setup.sh
```diff
@@ -26,16 +27,21 @@ setup_repo() {
...
     if (( in_repo )); then
-        cd "$meta_dir" || exit
+        cd "$main_dir" || exit
         git checkout "$BRANCH"
     else
         if [[ ! -d "$PWD/bitbots_main" ]]; then
-            git clone "$REPO_URL"
+            echo "Cloning repository bitbots_main..."
+            # Try to clone via SSH first, fall back to HTTPS if that fails
+            if ! git clone "$REPO_URL_SSH"; then
```
Flova: I am not sure if I like this implicit fallback. It might cause problems later on if people try to push via ssh.

scripts/setup.sh
```diff
@@ -45,14 +51,21 @@ setup_repo() {
...
 }
 
 build_repository() {
-    echo "Running full colcon build..."
+    echo "Running full build..."
     set +u
+
+    # Append "--packages-skip bitbots_basler_camera" to the build command if setup_host was skipped or failed
+    if (( basler_installed )); then
     $HOME/.pixi/bin/pixi run build
```
texhnolyze: Indentation

scripts/setup.sh
```diff
@@ -45,14 +51,21 @@ setup_repo() {
...
+    else
+    $HOME/.pixi/bin/pixi run build --packages-skip bitbots_basler_camera
```
texhnolyze: Indentation

scripts/setup.sh
texhnolyze:
```suggestion
setup_pixi() {
    if ! type pixi &> /dev/null; then
        curl -fsSL https://pixi.sh/install.sh | sh
    fi
}
```
This prevents unnecessary downloads and overwriting a global pixi install (e.g. on arch) with a local one.

texhnolyze (changes requested):
Does it make sense to add a tiny paragraph to old docs referencing,
`ros2 run ...` or `colcon build ...` commands directly 
that all ros/colcon commands should be done in a `pixi shell`?
Or do you think that's self explanatory?

scripts/make_basler.sh
```diff
@@ -7,6 +7,7 @@ set -eEo pipefail
 # Go to the download button and copy the link address.
 PYLON_DOWNLOAD_URL="https://data.bit-bots.de/pylon_7_4_0_14900_linux_x86_64_debs.tar.gz.gpg"
```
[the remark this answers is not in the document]
jaagut: I don't think we should upgrade, as we don't intend to use Wolfgang with its Basler camera anymore.

texhnolyze (approved):

PR #742, opened 2026-01-30
Bump deps

Flova:
# Summary

Bump deps to include the newest fixes from robostack. Not tested in sim yet.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -12,8 +12,7 @@ environments:
       pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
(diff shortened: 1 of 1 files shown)


jaagut:
Previously with requirements.txt, we had a frozen protobuf dependency because of mimic. But I guess this is fine since we need to replace mimic anyway and wolfgang has no functioning speakers.

Flova:
I think mimic is not even installed with pixie right now and the protobuff dep is for the team comm

PR #754, opened 2026-02-25
Use supertone tts

### LINKED ISSUE #752: Reintegrate TTS
# Summary
There was some work in the past to reintegrate a TTS mechanism: https://github.com/bit-bots/bitbots_main/pull/686

But @Flova has a suggestion for a better and more performant TTS system, which we want to integrate instead at the German Open.

This will also be used on the X02 in the future as it also has a speaker.

ayin21:
# Summary
Adds a fast and expressive diffusion based tts.

https://github.com/supertone-inc/supertonic

While it is available on pypi, it is mainly the helper.py file and the pypi package depends on onnxruntime, which breaks our dep on onnxruntime-webgpu. Therefor the file is just copied into our repo. The weights are pulled via our conda channel, which gets them from a huggingface git lfs.

## Related issues
Closes #752 

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


src/bitbots_misc/bitbots_tts/bitbots_tts/tts.py
```diff
@@ -96,7 +98,16 @@ def run_speaker(self) -> None:
             # Get the next message and speak it
             text, _ = self.prio_queue.pop(0)
             try:
-                say(text)
+                with timer("TTS Generation Time"):
+                    wav_untrimmed, duration = self.generate_speech(f"{text}")
+                wav = wav_untrimmed[0, : int(self.text_to_speech_engine.sample_rate * duration[0].item())]
+                wav = np.concatenate([np.zeros(2000), wav])
```
jaagut: What is this for?

Flova: I otherwise encountered an issue where the beginning was cut off because the soundcard was in a sleep state and needed a moment to wake up if there was no audio playing for a few seconds. Adding a bit of silence in the beginning fixed it for my PC.

jaagut (approved):

PR #756, opened 2026-03-08
fix(bitbots_ros_control): dynamixel driver include

texhnolyze:
to point to `<dynamixel_workbench_toolbox/dynamixel_driver.h>` the
correct include path for the `pixi` environment and not a local/global
installation coming with the dynamixel workbench GUI.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


Flova:
Why did this work without it? In e.g. the ci only the pixie headers are present.

PR #767, opened 2026-03-12
Feature/hsl gamecontroller migration

texhnolyze:
# Summary
I've tried to extract, clean up and condense the game controller changes...

And there still is some stuff I'd like to improve ideally, before merging (i.e. remove TODOs, don't make HCM depend on behavior for game controller, move publishing of return data to central place and maybe create a full custom Message for the whole return object to the gamecontroller in https://github.com/bit-bots/game_controller_hsl reducing the number of pub/subs one needs).

**We will just do the fixes/changes at a later stage, after the competition (see comments).**

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/decisions/fallen.py
```diff
@@ -23,32 +24,43 @@ def __init__(self, blackboard, dsd, parameters):
...
         if fused_roll > self.fallen_orientation_thresh:
+            self.publish_if_fallen(True)
             return "FALLEN_RIGHT"
 
         if fused_roll < -self.fallen_orientation_thresh:
+            self.publish_if_fallen(True)
             return "FALLEN_LEFT"
 
+        self.publish_if_fallen(False)
         return "NOT_FALLEN"
```
Flova: This is already covered by the HCM robot state

src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/decisions/fallen.py
```diff
@@ -23,32 +24,43 @@ def __init__(self, blackboard, dsd, parameters):
     def perform(self, reevaluate=False):
         # Check of the fallen detection is active
         if not self.blackboard.is_stand_up_active:
+            self.publish_if_fallen(False)
```
texhnolyze: Apparently there is a `RobotControlState` where we could read the info if we are fallen from, not requiring extra logic here.

PR #804, opened 2026-04-12
Assume the volume of goal post while decide to kick straight

### LINKED ISSUE #729: Goal orientation switches to straight shot despite goal post being in the way
# Summary
When trying to shoot a goal the robot does a straight shot instead of an angle one. This means the ball hits the goal post and doesn't go into the goal despite such a shot being easily possible.

## Expected behavior
Do an angle shot to avoid the goal post.

## Current behavior
Does a straight shot while the ball is position straight in front of the pole.

## Steps to Reproduce

1. Get the robot into a playing game state.
2. Position the ball in front of either of the posts such that the midpoint of the ball is inside the bounds defined by the midpoints of the goal posts.
3. Let the robot take a shot.
4. The ball hits the post and doesn't make it into the goal.

Image

## Possible Solution

Presumably this is caused by the algorithm checking if the ball is between the insides of the midpoints of the posts, not taking into account the physical extend of the posts. This could be fixed by either narrowing the area of the goal to match the insides of the posts or to take into account the physical extend.

cleWu03:
# Summary

In the decision if we are direct infront of the goal we now subtract 6cm (max width of goal post is 12 cm, means 6 cm from the defined center in the config)
## Related issues

From Issue #729

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/pathfinding_capsule.py
+++ b/src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/pathfinding_capsule.py
@@ -160,6 +160,7 @@ class PathfindingCapsule(AbstractBlackboardCapsule):
             ball_x, ball_y = self._blackboard.world_model.get_ball_position_xy()
 
             # Play in any part of the opponents goal, not just the center
+            # Adjust for the goal post width (-0.06)
             if abs(ball_y) < self._blackboard.world_model.goal_width / 2 - 0.06:
                 goal_angle = 0
             # Play in the opposite direction if the ball is near the opponent goal back line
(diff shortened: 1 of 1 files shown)


src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/pathfinding_capsule.py
```diff
@@ -160,7 +160,7 @@ def get_ball_goal(self, target: BallGoalType, distance: float, side_offset: floa
             ball_x, ball_y = self._blackboard.world_model.get_ball_position_xy()
 
             # Play in any part of the opponents goal, not just the center
-            if abs(ball_y) < self._blackboard.world_model.goal_width / 2:
+            if abs(ball_y) < self._blackboard.world_model.goal_width / 2 - 0.06:
```
jaagut, line 164:
```suggestion
            # Adjust for the goal post width (-0.06)
            if abs(ball_y) < self._blackboard.world_model.goal_width / 2 - 0.06:
```

jaagut (approved):

PR #854, opened 2026-05-12
Add Noise to the IMU sensors in the Mujoco simulation

### LINKED ISSUE #825: add imu noise to simulation
See look at the imu noise on the real robot. Than add some noise to the simulation for more realistic testing

MegaIng:
# Summary
Adds noise for the IMU in the Mujoco simulation. The default parameters approximate a simple measurement of a real robot, but no systematic measurement has been done. No systematic biases have been measured, which matches others experiences. The ability to add these has still been provided incase people want to test this ability out.

## Proposed changes
Add helper classes `NoisySensor` and `NoisyQuternionSensor`. This would also simplify if we ever want to add noise to the other sensors.

Add parameter handling for the noise parameters. Also update handling of a few of the older parameters, AFAICT, the `get_parameter_or` does not actually work the way it was used before (at least I couldn't get it to work with my new parameters).

## Related issues
closes #825 

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] ~~Test on the robot~~


src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/robot.py
```diff
@@ -1,20 +1,44 @@
...
         self.index: int = index
+        self._rng = rng or np.random.default_rng()
+        self.noise_config = noise_config or {}
         self.camera = Camera(model, data, name=self._get_name("head_camera"))
 
         def j(ros_name: str) -> Joint:
             return Joint(model, data, ros_name=ros_name, name=self._get_name(ros_name))
 
+        def s(base_name: str, ros_name: str) -> Sensor:
+            return Sensor(model, data, name=self._get_name(base_name), ros_name=ros_name)
```
jaagut: Is it correct, that `_get_name()` is called with  `base_name`? Other cases around here are different.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
```diff
@@ -27,6 +32,64 @@ def dim(self) -> int:
...
+        name: str,
+        ros_name: str,
+        gaussian_stddev: float = 0.0,
+        bias_stddev: float = 0.0,
+        rng: np.random.Generator | None = None,
+        fixed_bias: np.ndarray | None = None,
+    ):
+        super().__init__(model, data, name, ros_name)
+        self.gaussian_stddev = float(gaussian_stddev)
+        self.bias_stddev = float(bias_stddev)
```
jaagut:
```suggestion
        self.gaussian_stddev = gaussian_stddev
        self.bias_stddev = bias_stddev
```

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
```diff
@@ -27,6 +32,64 @@ def dim(self) -> int:
...
+    @property
+    def noisy_data(self) -> np.ndarray:
+        """Gets the sensor reading with fixed bias and additive Gaussian noise."""
+        return self.data + self._fixed_bias + self._rng.normal(scale=self.gaussian_stddev, size=self.dim)
+
+    def reset_fixed_bias(self, fixed_bias: np.ndarray | None = None) -> None:
+        """Resets the fixed bias for this sensor."""
+        if fixed_bias is not None:
+            self._fixed_bias = np.asarray(fixed_bias, dtype=float)
+        elif self.bias_stddev > 0:
```
jaagut, line 70: `fixed_bias` is an array. This check should fail with:

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
```diff
@@ -16,6 +16,11 @@ def data(self) -> np.ndarray:
         """Gets the current sensor reading from sensordata array."""
         return self._data.sensordata[self.adr : self.adr + self.dim]
 
+    @property
+    def noisy_data(self) -> np.ndarray:
+        """Gets the current sensor reading with measurement noise applied."""
+        return self.data
+
```
jaagut, line 23: I am not so happy with always calling `noisy_data` even for non-noisy Sensors. I would do the reverse, if there is no reason to get the raw non-noisy data of a NoisySensor.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
```diff
@@ -27,6 +32,64 @@ def dim(self) -> int:
         return int(self.instance.dim)
 
 
+class NoisySensor(Sensor):
+    """Represents a sensor with per-start fixed bias and additive Gaussian noise."""
```
jaagut: What do you mean with `per-start`?

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/robot.py
```diff
@@ -48,14 +72,22 @@ def j(ros_name: str) -> Joint:
...
-                Sensor(model, data, name=self._get_name("accelerometer"), ros_name="IMU_accelerometer"),
-                Sensor(model, data, name=self._get_name("orientation"), ros_name="IMU_orientation"),
-                Sensor(model, data, name=self._get_name("position"), ros_name="IMU_position"),
-                Sensor(model, data, name=self._get_name("l_foot_pos"), ros_name="left_foot_position"),
-                Sensor(model, data, name=self._get_name("r_foot_pos"), ros_name="right_foot_position"),
-                Sensor(model, data, name=self._get_name("l_foot_global_linvel"), ros_name="left_foot_velocity"),
-                Sensor(model, data, name=self._get_name("r_foot_global_linvel"), ros_name="right_foot_velocity"),
+                n("gyro", "IMU_gyro", "gyro_gaussian_stddev", "gyro_bias_stddev"),
+                n("accelerometer", "IMU_accelerometer", "accelerometer_gaussian_stddev", "accelerometer_bias_stddev"),
+                QuaternionNoisySensor(
```
jaagut: Please also add a helper function for this.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/robot.py
```diff
@@ -1,20 +1,44 @@
...
+    ):
         self.index: int = index
+        self._rng = rng or np.random.default_rng()
+        self.noise_config = noise_config or {}
         self.camera = Camera(model, data, name=self._get_name("head_camera"))
 
         def j(ros_name: str) -> Joint:
             return Joint(model, data, ros_name=ros_name, name=self._get_name(ros_name))
 
+        def s(base_name: str, ros_name: str) -> Sensor:
```
jaagut: At this point, these one-letter methods (even without docs) are not helpful. Please award them a more descriptive name.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
MegaIng: I can imagine usecases for getting the non-noisy data, e.g. when providing data for my automated testing framework developed in the context of StuFo (which is currently still on wolfgang and I wont try to merge it till StuFo is done).

But that can be done just as well using `raw_data` & `data` instead of `data` & `noisy_data`. I don't have a strong preference between the two, keeping `data` intact as the groundtruth felt less disruptive and more predictable to me. If you want it the other way around I can do it.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
MegaIng: "per-start" meaning per simulator startup, i.e. everytime these instances are created. I agree it's not the proper formulation, I am going with a per-instance formulatio instead.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/robot.py
MegaIng: `Joint`'s use `ros_name == base_name`. So yes this is correct

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
MegaIng: You misread the code, we are checking `self.bias_stddev` which is a float, not an array. The above check `fixed_bias is not None` is not a boolean check either.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/sensor.py
jaagut: Okay, now I agree more with how you already did it.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
jaagut: You actually use this in line 66, so it will fail if `TYPE_CHECKING == False`
```suggestion
from bitbots_mujoco_sim.robot import Robot
from bitbots_mujoco_sim.simulation import RobotSimulation
```

jaagut: Hmm. GitHub's suggestions are broken.

MegaIng: This is not something I changed. This is also correct since `RobotSimulation` actually gets defined in this file.

I am less sure why this is forward declared in this way.

jaagut (approved):

PR #832, opened 2026-05-11
Framework/rl standup 

BloodWork44:
# Summary
add rl_standup node an handlers to the  rl_framework for deploying the standup policies 

## Related issues
closes #814 

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot
- [ ] Create issues for future work
- [ ] Triage this PR and label it

PR #858, opened 2026-05-13
Feature/rl framework integration

jgueldenstein:
# Summary
integrate the mujoco playground trained walk in the bitbots_rl_motion framework

includes some changes from #832 

## Checklist

- [x] Run `pixi run build`
- [x] Test on your machine


src/bitbots_robot/piplus_description/urdf/pi_plus_22dof.urdf.xacro
jaagut, line 85: This has just been changed yesterday in https://github.com/bit-bots/bitbots_main/pull/842

What is this?

src/bitbots_misc/bitbots_bringup/launch/motion_standalone.launch
```diff
@@ -3,7 +3,7 @@
     <arg name="sim" default="false"/>
     <arg name="viz" default="false"/>
     <arg name="torqueless_mode" default="false" description="start without torque, for example for testing the falling detection"/>
-    <arg name="tts" default="true" description="Whether to enable text-to-speech"/>
+    <arg name="tts" default="false" description="Whether to enable text-to-speech"/>
```
Flova: Why is this disabled now?

src/bitbots_misc/bitbots_bringup/launch/motion.launch
```diff
@@ -2,9 +2,8 @@
 <launch>
     <arg name="sim" default="false"/>
     <arg name="viz" default="false"/>
-    <arg name="walking" default="true" description="start walking" />
     <arg name="torqueless_mode" default="false" description="start without torque, for example for testing the falling detection"/>
-    <arg name="tts" default="true" description="Whether to speak" />
+    <arg name="rl_motion" default="true" description="Whether to use the RL motion system instead of the regular one" />
```
Flova, line 6: What is the regular one in this case. It seems to replace the other one.

src/bitbots_misc/bitbots_bringup/launch/simulator_teamplayer.launch
```diff
@@ -11,6 +11,7 @@
     <arg name="teamcom" default="false" description="Whether the team communication system should be started" />
     <arg name="vision" default="true" description="Whether the vision system should be started" />
     <arg name="world_model" default="true" description="Whether the world model should be started"/>
+    <arg name="rl_motion" default="true" description="Whether to use the RL motion system instead of the regular one" />
```
Flova, line 14: Same here.

src/bitbots_motion/bitbots_rl_motion/handlers/joint_handler.py
```diff
@@ -13,57 +13,63 @@ def __init__(self, node):
...
             )
             - self._walkready_state
         )
 
         return joint_angles
 
-    def get_velocity_data(self):
+    def get_velocity_data(self) -> np.ndarray:
         assert self._joint_state is not None
+        if self._joint_state_indices is None:
```
Flova, line 60: This my might be dangerous in some situations if e.g. some joint states are missing later on and the array gets shifted. Probably it's fine with the way we communicate with the motors, but I don't think the optimization safes that much compared to a potentially hard to debug issue.

src/bitbots_motion/bitbots_rl_motion/nodes/mjlab_walk_node.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+from handlers.ball_handler import BallHandler
```
Flova, line 2: Is this needed for the mjlab walk?

src/bitbots_motion/bitbots_rl_motion/nodes/walk_node.py
```diff
@@ -49,8 +49,23 @@ def publisher(self, onnx_pred):
 
     # states in which the policy executes
     def allowed_states(self):
-        allowed_to_move = self._robot_state_handler.is_walkable() and np.any(self._command_handler.get_command() != 0.0)
-        return allowed_to_move
+        return self._robot_state_handler.is_walkable()
```
Flova, line 52: We might need to think a bit about the robot states and how we integrate the kick. Also how we trigger a state change in the HCM for the kick. But we need to think about the kick interface for that. A topic based solution seems fine, but I have some concerns on how we go back to the walking after the kick for example. Especially how exactly we integrate this with the hcm.

src/bitbots_robot/piplus_description/urdf/pi_plus_22dof.urdf.xacro
jgueldenstein: The same changes

src/bitbots_misc/bitbots_bringup/launch/motion.launch
jgueldenstein: the tts should not be part of the motion launch file,
walking is renamed to rl motion as the rl motion node does not only do walking but possibly also kicking and standing up in the future

src/bitbots_misc/bitbots_bringup/launch/motion_standalone.launch
jgueldenstein: because the robot said too many things during testing

src/bitbots_motion/bitbots_rl_motion/handlers/joint_handler.py
jgueldenstein: adressed by 7445a27f2

src/bitbots_motion/bitbots_rl_motion/nodes/mjlab_walk_node.py
jgueldenstein: probably not, the mjlab files are not ready yet, they will be added in a separate pr

src/bitbots_motion/bitbots_rl_motion/nodes/walk_node.py
jgueldenstein: True, good to keep in mind for the future, but this is a required change because otherwise the robot stops sending joint commands once the velocity command is 0, leading to falls.

src/bitbots_misc/bitbots_bringup/launch/motion_standalone.launch
jgueldenstein: reenabled in [e7c45d5](https://github.com/bit-bots/bitbots_main/pull/858/commits/e7c45d5925583ac38dd925ce1995614d30f2e8f4)

jgueldenstein:
Tested on real robot with motion standalone.

Flova (approved):

PR #860, opened 2026-05-17
Odometry with better contact model

Flova:
# Summary

Reworking the odometry from scratch to include a better phase detection, detailed foot contact modeling and fused imu data.

## Proposed changes

- Drop odometry fuser
- Determine phase inside the odometry, considering imu data and propreoception to determine which contact point on which foot is in contact
- Model movement around contact points to better handle "rolling motions" during walking
- Integrate odometry fuser into joined node to rotate the robot around this now very detailed contact point estimate 

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


src/bitbots_navigation/bitbots_odometry/src/motion_odometry.cpp
```diff
@@ -1,181 +1,381 @@
...
+      l_back_right_corner_frame_;
+  rclcpp::Publisher<nav_msgs::msg::Odometry>::SharedPtr pub_odometry_;
+  rclcpp::Publisher<biped_interfaces::msg::Phase>::SharedPtr pub_support_state_;
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub_;
+  sensor_msgs::msg::Imu::SharedPtr imu_data_;
+
+  motion_odometry::ParamListener param_listener_;
+  motion_odometry::Params config_;
+
+  // Returns {contact_corner_frame, sole_frame} of the foot with the lowest corner in IMU world space.
```
MegaIng: This comment is not accurate, lowest corner for both foots are returned right now.

src/bitbots_navigation/bitbots_odometry/src/motion_odometry.cpp
```diff
@@ -1,181 +1,381 @@
...
+  if (lowest_z_l < lowest_z_r) {
+    support_state = biped_interfaces::msg::Phase::LEFT_STANCE;
+  } else {
+    support_state = biped_interfaces::msg::Phase::RIGHT_STANCE;
+  }
+
+  return {lowest_corner_l, lowest_corner_r, support_state};
+}
+
+tf2::Transform MotionOdometry::getFootContactModelResult(const rclcpp::Time& time,
```
MegaIng: Since the time parameter is currently being ignored, it should not be part of the signature. Alternatively keep the parameter, use it and set it to 0 externally with explanation.

src/bitbots_navigation/bitbots_odometry/src/motion_odometry.cpp
```diff
@@ -1,181 +1,381 @@
...
+    tf2::Transform rotation_point_in_base_link;
+    tf2::fromMsg(rotation_point_in_base_link_msg.transform, rotation_point_in_base_link);
+
+    tf2::Transform imu_mounting_offset;
+    try {
+      geometry_msgs::msg::TransformStamped imu_mounting_transform =
+          tf_buffer_.lookupTransform(imu_frame_, base_link_frame_, rclcpp::Time(0));
+      fromMsg(imu_mounting_transform.transform, imu_mounting_offset);
+    } catch (tf2::TransformException& ex) {
+      RCLCPP_ERROR(this->get_logger(), "Not able to fuse IMU data with odometry due to a tf problem: %s", ex.what());
```
MegaIng, line 225: Why is this error state specifically protected? Also, isn't this an essentially static lookup that just depends on the URDF? Under what conditions might this fail?

src/bitbots_navigation/bitbots_odometry/src/motion_odometry.cpp
```diff
@@ -1,181 +1,381 @@
...
+
+    // Apply the walked amount to the current state
+    // Go from odom to base_link, then to the rotation point and apply the movement since the last time there
+    tf2::Transform odom_to_rotation_point = odom_to_base_link_ * rotation_point_in_base_link * previous_to_current;
+
+    // Go back to the base_link frame as the odom is defined between odom and base_link
+    odom_to_base_link_ = odom_to_rotation_point * rotation_point_in_base_link.inverse();
+
+    // Just to be sure, we set the rotation to the current imu orientation
+    odom_to_base_link_.setRotation(imu_orientation_in_base_link);
```
MegaIng, line 246: Am I reading this correctly that we always use the IMU rotation and trust it without any input from the odometry? Not necessarily bad, just want to make sure.

MegaIng (changes requested):

MegaIng (approved):

PR #805, opened 2026-04-12
team_com disallowed robots to send messages if not intended, and reduces the rate if we get close to the budget end

cleWu03:
# Summary

- Reduce team com rate if we are getting close to the limit (defenition of "close" depends on the time resuming)
- Stop sending team com msg if we are very close to the limit ("very close" = 40 msg left, so with delay we can still send 10 msg per robot)
- NOT ready!!! for merge, we need to check effects on behavior first

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot
- [ ] Create issues for future work
- [ ] Triage this PR and label it

PR #870, opened 2026-05-24
team_com disallowed robots to send messages if not intended, and reduces the rate if we get close to the budget end

### LINKED ISSUE #748: Adjust Team Communication to new rules
# Summary
According to the new rules, we are maximally allowed to send 12000 messages per Team, but are currently running the team communication with 10Hz per robot.
This means we send 12000 messages per robot playing per game (20min).

So we will need to reduce the number of messages we send and then see how that affects behavior and adjust it accordingly.

There is also a maximum size of messages, which we have to check our `protobuf` message against.

MegaIng:
See https://github.com/bit-bots/bitbots_main/pull/805

Closes #748

@cleWu03  @jaagut

---
<a href="https://app.blacksmith.sh/bit-bots/codesmith/bitbots_main/pr/870"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782225443&installation_id=134975672&pr_number=870&repository=bit-bots%2Fbitbots_main&return_to=https%3A%2F%2Fgithub.com%2Fbit-bots%2Fbitbots_main%2Fpull%2F870&signature=7c6124ba47a360cfbce50c8a8065402a77cecde475c01821f4442491e674d72b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>


src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -275,8 +282,11 @@ def is_still_valid(time: Optional[TimeMsg]) -> bool:
 
         message = self.protocol_converter.convert_to_message(self, msg, is_still_valid)
         proto_msg = message.SerializeToString()
-        self.logger.debug(f"Sending msg with size {len(proto_msg)} bytes")
-        self.socket_communication.send_message(proto_msg)
+        if(len(proto_msg) > 512):
+            self.logger.warning(f"Team_com msg not sended, because size {len(proto_msg)} bytes is above the maximum of 512 bytes")
```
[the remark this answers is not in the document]
jaagut: Please don't hardcode 512, but put it into a configurable parameter. Use this variable in the warning f-string.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -313,7 +331,29 @@ def extract_orientation_yaw_angle(self, quaternion: Quaternion):
...
+        # once our actual msg budget is below this linear rate we tend to send more msg then allowed and should reduce our sending rate
+        if self.game_started_recently():
+            return False
+        
+        msg_left_linear_rate = (self.gamestate.first_half * 600 + self.gamestate.secs_remaining) * 4 * 2.5
+        return msg_left_linear_rate > self.gamestate.message_budget
+    
+    def game_started_recently(self):
+        #true in the first 60 seconds of the game
+        return self.gamestate.first_half and self.gamestate.secs_remaining > 540
```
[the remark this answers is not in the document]
jaagut: We should use the helpers

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -301,7 +311,15 @@ def should_message_be_discarded(self, message: Proto.Message) -> bool:
...
     def is_robot_allowed_to_send_message(self) -> bool:
-        return self.gamestate is not None and not self.gamestate.penalized
+        #a penalized robot doesn't need to publish
+        if self.gamestate is not None and not self.gamestate.penalized:
+            return False
+        #if we are close to our message budget, we dont want to continue publishing
+        #the budget smaller 40 as stop definition makes sure we have 10 msg per robot left in case of some delay in the communication with the game controller 
+        if self.gamestate is not None and (self.gamestate.message_budget < 40):
+            return False
+
```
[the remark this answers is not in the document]
jaagut: This needs a human discussion first.

jaagut:
@copilot please resolve the found issues. Omit the comment that still needs discussion. At the end, run `pixi run -e format format` to make sure the code complies.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -313,7 +336,41 @@ def extract_orientation_yaw_angle(self, quaternion: Quaternion):
...
+        half_duration_seconds = 600
+        team_size = 4
+        msgs_per_second_per_robot = 2.5
+
+        # We are allowed to send 2.5 messages per second on average with each robot (12000 with 4 robots in a 1200 seconds game).
+        # the msg_left_linear_rate is the amount of messages we would send if we send exactly with this 2.5 msg per sec per robot rate
+        # once our actual msg budget is below this linear rate we tend to send more msg then allowed and should reduce our sending rate
+        if self.game_started_recently():
```
jaagut, line 358: @cleWu03 why is this necessary?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -145,6 +156,7 @@ $IsPenalized
     JUST_UNPENALIZED --> $GameStateDecider
         INITIAL --> #Init
         ELSE --> #GetWalkreadyAndLocalize
+    UNPENALIZED_AFTER_TEAM_COM_LIMIT --> #DoNothing
```
jaagut, line 150: What's the thinking here? Why are we passive after being unpenalized? Is there a possibility this hits every team robot at the same time frame and thus disables the whole team while still being allowed to play?

jaagut (approved):

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
cleWu03: Since the unpenalised robot won't receive any more team_com messages it would act like it's playing alone and go to the ball. This could hinder the actual striker while playing. The only case the whole team stops playing is where all robots are penalised at the moment the team com_limit is reached, since a penalised robot is not allowed to send any more messages this is a very very rare case

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
cleWu03: To handle variances in the data at the beginning of a game

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -195,7 +221,9 @@ def is_not_goalie(team_data: TeamData) -> bool:
             team_data_infos = filter(is_not_goalie, team_data_infos)  # type: ignore[assignment]
 
         # Count valid team data infos (aka robots with valid team data)
-        return sum(map(self.is_valid, team_data_infos))
+        active_players = sum(map(self.is_valid, team_data_infos))
+        self.last_number_of_active_players = active_players
```
Flova, line 236: This is ugly because the side effect on the function is not obvious. It might get called at different point with different parameters and the stored state will differ based on that.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -209,7 +237,9 @@ def is_a_goalie(team_data: TeamData) -> bool:
         team_data_infos = filter(is_a_goalie, team_data_infos)  # type: ignore[assignment]
 
         # Count valid team data infos (aka robots with valid team data)
-        return sum(map(self.is_valid, team_data_infos)) == 1
+        goalie_count = sum(map(self.is_valid, team_data_infos))
+        self.was_goalie_active = goalie_count == 1
```
Flova: Same here. It is not set if we don't call the function at some point. Make it instead its own function that checks the last team com message that we have got.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -28,9 +28,14 @@ def __init__(self, blackboard, dsd, parameters):
         super().__init__(blackboard, dsd, parameters)
 
     def perform(self, reevaluate=False):
-        my_time_to_ball = self.blackboard.team_data.get_own_time_to_ball()
-        rank = self.blackboard.team_data.team_rank_to_ball(my_time_to_ball, count_goalies=True, use_time_to_ball=True)
-        self.publish_debug_data("time to ball", my_time_to_ball)
+        if self.blackboard.gamestate.get_team_com_limit_has_reached():
+            rank = self.blackboard.team_data.get_last_rank_with_team_com()
```
Flova, line 32: I am not sure if I like how much explicit behavior is added for this edge case. It should not happen. And if it does happen it is okay if we play with less robots for example. There is a lot of potential for bugs and maintaining overhead with this implementation.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -47,9 +52,14 @@ def __init__(self, blackboard, dsd, parameters):
         super().__init__(blackboard, dsd, parameters)
 
     def perform(self, reevaluate=False):
-        my_time_to_ball = self.blackboard.team_data.get_own_time_to_ball()
-        rank = self.blackboard.team_data.team_rank_to_ball(my_time_to_ball, count_goalies=False, use_time_to_ball=True)
-        self.publish_debug_data("time to ball", my_time_to_ball)
+        if self.blackboard.gamestate.get_team_com_limit_has_reached():
+            rank = self.blackboard.team_data.get_last_rank_with_team_com()
```
Flova: There is also the chance that two robots get assigned the same role with this setup if we have an unlucky moment. If we want to handle this we could just switch to a setup based on the initial roles. They are well defined and don't need much extra handling.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/goalie_active.py
```diff
@@ -13,10 +13,16 @@ def __init__(self, blackboard, dsd, parameters):
         super().__init__(blackboard, dsd, parameters)
 
     def perform(self, reevaluate=False):
-        if self.blackboard.team_data.get_is_goalie_active():
-            return "YES"
+        if self.blackboard.gamestate.get_team_com_limit_has_reached():
+            if self.blackboard.team_data.get_was_goalie_active():
```
Flova, line 17: Again I would prefer a hard default state.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -120,14 +122,23 @@ $BallSeen
...
             GOALIE --> #GoalieBehavior
-            ELSE --> $CountActiveRobotsWithoutGoalie
-                ONE --> $RankToBallNoGoalie
-                    FIRST --> #StrikerRole
-                    SECOND --> #DefensePositioning
-                ELSE --> $RankToBallNoGoalie
-                    FIRST --> #StrikerRole
-                    SECOND --> #SupporterRole
-                    THIRD --> #DefensePositioning
+            ELSE --> $TeamComLimitReached //this decision is only for better visualization in rqt
```
Flova: Please do a dsd debug log instead. It is also visualized that way and it does not clutter the dsd that way.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
Flova: This would not be an issue if we play with a predefined setup. We could even vary it a bit and play with e.g. 3 defenders 1 striker and a goalie, as the supporter is pretty useless.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -301,7 +314,17 @@ def should_message_be_discarded(self, message: Proto.Message) -> bool:
         return is_own_message or is_message_from_oposite_team
 
     def is_robot_allowed_to_send_message(self) -> bool:
-        return self.gamestate is not None and not self.gamestate.penalized
+        
+        if self.gamestate is not None:
+            #a penalized robot doesn't need to publish
+            if self.gamestate.penalized:
+                return False
```
Flova: Not a great pattern to return true or false based on an if. You can just return the result of the Boolean operation directly.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -301,7 +314,17 @@ def should_message_be_discarded(self, message: Proto.Message) -> bool:
...
+            #if we are close to our message budget, we dont want to continue publishing
+            #the budget smaller 40 as stop definition makes sure we have 10 msg per robot left in case of some delay in the communication with the game controller 
+            if self.gamestate.message_budget < 40:
+                return False
+        
+        return True
```
Flova: This seems wrong. Previously we did not send if game state was none now we do.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -313,7 +336,41 @@ def extract_orientation_yaw_angle(self, quaternion: Quaternion):
...
         return transforms3d.euler.quat2euler([quaternion.w, quaternion.x, quaternion.y, quaternion.z])
-
+    
+    def reduce_rate(self):
+        self.timer.cancel()
+        self.create_timer(self.reduced_rate)
+        self.rate_is_reduced = True
+        self.logger.warning("Team communication: message sending rate is reduced now")
+
+    def create_timer(self, rate: int):
```
Flova, line 342: Are we sure the previous timer is stopped? Was this tested?

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
Flova: What do you mean with that?

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -313,7 +336,41 @@ def extract_orientation_yaw_angle(self, quaternion: Quaternion):
...
+        team_size = 4
+        msgs_per_second_per_robot = 2.5
+
+        # We are allowed to send 2.5 messages per second on average with each robot (12000 with 4 robots in a 1200 seconds game).
+        # the msg_left_linear_rate is the amount of messages we would send if we send exactly with this 2.5 msg per sec per robot rate
+        # once our actual msg budget is below this linear rate we tend to send more msg then allowed and should reduce our sending rate
+        if self.game_started_recently():
+            return False
+
+        msg_left_linear_rate = (
```
Flova, line 361: We might want to smooth this

Flova: Maybe calculated over a window of a minute.

Flova: Also I am not sure how much I like the reduced rate feature, as it heavily affects team com dynamics. I would use one config up until a hard cutoff.

src/lib/game_controller_hsl/game_controller_hsl/scripts/sim_gamestate.py
```diff
@@ -116,6 +126,10 @@ def loop(self):
                     game_state_msg.game_phase = ord(key) - ord("a")
                 elif key in [chr(ord("e") + x) for x in range(7)]:
                     game_state_msg.set_play = ord(key) - ord("e")
+                elif key == "m":
```
Flova, line 129: I like that.

Flova (changes requested):

Flova:
Sorry if I am a bit strict with the review...

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
cleWu03: I am strictly against using the initial roles, since we assigne the roles dynamic during the game the possibility of having much unlucky role switching is way higher then the chance of problems with this implementation.  E.g. we have two players left, goalie is handling the ball and close to the opponents goal, while the configured striker has the goalie role. In this case it would be terrible to let the goalie walk back inside our own goal when he has almost scored a goal. Since the chance of reaching the limit is very low it should only occur in the last seconds of a game, so my aim is to continue the on going action, instead of having a stable solution for a longer period of play

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
cleWu03: I made new functions for all saves of further team com data, to make sure they are identical with the current dsd decisions they are used  inside there, instead of checking a old team com msg

Flova: I am not fully sure what you man by this. Maybe we need to talk about it in person, but these side effects are a big foot gun.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
Flova: I think this feature is a last resort either way and if it triggers something probably went wrong at some point. If this is the case it could also happen earlier. While I agree with you that swapping roles is not the best behavior, it is the most reliable and simple one. Especially if we play for an extended period because something went wrong earlier. But we need to define our goals with it. Should it be a last resort that just stops us from doing nothing useful in an unlikely event or should it be smoothly integrated with a significantly higher complexity? I would argue that our new robots are reasonably fast and we should keep the implementation as simple and safe as possible even if it gives us a disadvantage for 15 seconds or so. The biggest disadvantage would be a behavioral bug that breaks everything in a situation like this. This is especially the case for a feature like that that will realistically only be tested a few times before merging and then sit there in the codebase until it is needed, while the codebase evolves around it. Therefore adopting a keep it stupid simple approach is favorable imo.. I don't want to discourage cool behavior features btw. but I think there are some lower hanging fruits and code paths that are more frequently visited that have a lot of potential for improvements.

Flova:
I am a bit cautious with setting global state in decisions. I think it might be okay in this case, but we need to be careful. Why is it not possible to parse this information out of the global team data state that already exists. The team data capsule should retain the last team data message from each robot and we can act accordingly. I think some functions like get rank to ball might have a timeout for those msgs, but we could add a flag that disables the timeout.

jaagut:
Superseded by #938

PR #882, opened 2026-06-01
Add team player tui

ayin21:
Adds Teamplayer orchestration tui/CLI.

As this is build in rust I used to chance to do it properly and port/extend the bitbots_cargo that was forgotten on another branch. This allows us to also build the bitbots_rust_nav / now bitbots_global_planner in the normal colcon build process. Also ports the very simple extrinsic calibration node to rust to show the ROS bindings.


jaagut:
```
feature/teamplayer_tui  pixi run teamplayer
✨ Pixi task (teamplayer in default): ros2 run bitbots_bringup teamplayer: (Launch the main software components)
No executable found
```

Flova:
Clean build?

Flova:
I also pushed some fixes

jaagut:
Some usage feedback:
- The giant header should read "Bit-Bots" not "Bit Bots". Maybe add the logo (small, simplified?)?
- Why would I need to type in a different DSD file?
- Selecting "Simulator" puts the focus back to "Zenoh". So toggling that option requires noticing this and pressing `->` to go back.
- In Simulation, audio and whistle detection should be deselected by default.
- Can we have a drop-down for `Fieldname`?
- Focusing the "Launch" button is barely visible in my terminal.
- A node should not be displayed as "DONE" when it failed. 
  image
  image
- Pressing "s" to stop the motion does nothing or at least shows no logs. It should not show "Fail" but "Stopped", since we stopped it manually.
  image
- I associate the spinner animation with waiting for something to finish, thus I would expect it during the startup process. Healthy running components should have a more "confirmative" icon/animation.

jaagut:
I did not test this, but how does this behave, if some parts are already running? 

Can I pass arguments to select the start behavior. E.g. motion:=true audio:=false etc so we can run this tui inside tmux from the deploy tool to launch everything and have the nice functionality?

MegaIng:
IMO this would ideally be written in python, not rust. This is a piece of software that is not performance critical and may be adjusted regularly. I would also like to reuse part of this infrastructure for a sim-launcher TUI that would then delegate to multiple different team player tuis running for the different ros domains - I strongly suspect that this kind of reuse would be easier if this was written in python.

PR #885, opened 2026-06-04
fix sign in urdf r_ankle_pitch limit

jgueldenstein:
# Summary
check the l_ankle_pitch, where the limit is done correctly for reference


jaagut (approved):

Flova:
Is this also relevant for mujoco XML Files?

jgueldenstein:
I am working on a bit of an update to the xml files anyways to make the mjlab version and the ros sim version more similar

Issue #905, opened 2026-06-08
Animation Node does not set K_p and K_d


jgueldenstein:
# Summary
the animation node does not explicitly set the kp and kd values for animations.
this should be changed so that animations can be tuned with specific values. Standing up requires a higher kp in the arms than waving

## Possible Solution
set the kp and kd explicitly in the animation file, parse it and apply that to the robot, use a sensible default for other animations that do not have it set

Flova:
What would be the issue if the waving is executed with a high kp as long as the controller is stable? On Wolfgang we also did "normal" position control with fixed kp. The current policy was that the default pid controllers are for classic position control and lower ones can be set on demand dynamically. That being said for example for falling animations it could make sense to set them.

PR #900, opened 2026-06-07
Piplus dribble scoring

Flova:

PR #907, opened 2026-06-08
Fix/animation kd kp

jgueldenstein:
# Summary
Currently there is no way of specifying the kp and kd values for an animation.
see #905 

this pr changes that so that standing up may work more reliably and does not depend on the default kp and kd or previously set values in the hardware interface or in the simulation.

I have tested this in simulation and the robot does not fall anymore when starting motion standalone and it can reliably stand up from the front and back with the set values.


jgueldenstein:
This still has to be tested on hardware. Also we need to re-check if the set kp for the falling animation is appropriate as the head may be too stiff when falling.

jgueldenstein:
build and test fails due to #900 as far as I can tell

Flova:
I think the falling was mostly related to 1 second being a very tight timing from t pose to walkready in the sim. I would prefer spawning it in a different pose, sitting down or similar to walkready. I think fixing this with the kp is not a good solution. Or did I miss something. Also the robot was capable of a reliable standup in sim a few days ago, so there might be a different regression if there are issues now. Currently the animations are played with relatively stiff params as tracking accuracy is important for static animations. That being said I think there are cases for e.g. falling animations where low kp can make sense.

jaagut:
@Flova @jgueldenstein how do we progress with this?

ayin21 (approved):

Flova:
There is still an unresolved comment. And I was waiting for a reaction of @jgueldenstein . Please don't merge yet.

jgueldenstein:
closed as not planned, also closes #905

PR #913, opened 2026-06-10
feat: Implement dynamic target IP resolution and add corresponding tests

### LINKED ISSUE #578: Dynamically decide team communication IP

jaagut:
- Introduced `resolve_target_ip` function to determine target IP based on configuration or connected Wi-Fi interface.
- Updated `SocketCommunication` class to utilize dynamic IP resolution.
- Modified configuration to allow 'auto' setting for target IP.
- Added unit tests for target IP resolution and Wi-Fi interface detection.

# Summary

## Related issues

Closes #578

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
This needs testing on the robots

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/network.py
```diff
@@ -0,0 +1,76 @@
...
+
+def _has_carrier(interface_path: Path) -> bool:
+    """Check if the interface has a carrier (i.e., is connected). Returns False if the carrier file cannot be read."""
+    try:
+        return (interface_path / "carrier").read_text().strip() == "1"
+    except OSError:
+        return False
+
+
+def _get_ipv4_broadcast_address(interface_name: str) -> IPv4Address | None:
```
Flova, line 66: Isn't there a higher level library for this available?

Maybe netifaces?

jaagut: Maybe, but the netifaces project seems dead or unmaintained.

Flova (approved):

jaagut:
Thanks, this still needs testing

Flova:
Fyi one of the other PRs abstracts the communication layer in the team comm and adds a ROS backend in addition to the udp one. This enables the usage of multiple clients on one machine that communicate via a shared topics, which fits well into the way our multi robot mujoco setup is currently constructed.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/network.py
```diff
@@ -0,0 +1,74 @@
...
+
+def resolve_target_ip(configured_target_ip: str) -> tuple[IPv4Address, str | None]:
+    if configured_target_ip.strip().lower() != "auto":
+        return IPv4Address(configured_target_ip), None
+
+    return get_wifi_broadcast_address()
+
+
+def get_wifi_broadcast_address() -> tuple[IPv4Address, str]:
+    wifi_interfaces = [path for path in SYS_CLASS_NET.iterdir() if _is_wifi_interface(path)]
```
texhnolyze, line 22: Is the check for both "wireless" and "phy80211" required?
Otherwise this could be simplified to:
`[path.parent for path in Path(SYS_CLASS_NET).glob("*/wireless")]`

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/network.py
```diff
@@ -0,0 +1,74 @@
...
+        return False
+
+
+def _get_ipv4_broadcast_address(interface_name: str) -> IPv4Address | None:
+    for address in psutil.net_if_addrs().get(interface_name, []):
+        if address.family != socket.AF_INET or address.broadcast is None:
+            continue
+
+        broadcast_address = IPv4Address(address.broadcast)
+        return None if broadcast_address.is_unspecified else broadcast_address
```
texhnolyze, line 72: As far as I know `is_unspecified` only returns true if there is no actual specified address (meaning `0.0.0.0` for all interfaces), which I don't think can ever happen for a wifi interfaces with an actual connection.
Could we just directly return `IPv4Address(address.broadcast)`?

texhnolyze (approved):
Looks good to me. Only have two tiny adjustments if it makes sense 👍

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/network.py
jaagut: yes, this is necessary as this diffentiates between WiFi networks and others.

texhnolyze: But your code is:
```
def _is_wifi_interface(interface_path: Path) -> bool:
    return (interface_path / "wireless").is_dir() or (interface_path / "phy80211").exists()
```

And I'm confused `phy80211` is a WLAN standard and not a LAN, so the double check is unecessary, no?
For LAN devices the path `phy80211` does not exist.

PR #906, opened 2026-06-08
Mjmodel fixes

jgueldenstein:
# Summary
corrected joint ranges and actuator parameters for the mujoco model

PR #914, opened 2026-06-11
Update pi plus leg joint damping and armature params

Flova:
These params had a better sim to real transfer and where used for the latest policy training


jgueldenstein:
integrated in #906

PR #915, opened 2026-06-12
Add static obstacles to avoid running into the goals.

Flova:
## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/count_active_players_without_goalie.py
+++ b/src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/count_active_players_without_goalie.py
@@ -27,7 +27,7 @@ class CountActiveRobotsWithoutGoalie(AbstractDecisionElement):
             # emergency fall back if something goes wrong
             self.blackboard.node.get_logger().error("Rank to ball had some issues")
             raise ValueError(
-                f"Count active Robots encountered to many robots on the field: {number_of_active_teammates}"
+                f"Count active Robots encountered too many robots on the field: {number_of_active_teammates}"
             )
 
     def get_reevaluate(self):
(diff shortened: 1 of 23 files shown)


src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/count_active_players_without_goalie.py
```diff
@@ -26,7 +26,9 @@ def perform(self, reevaluate=False):
         else:
             # emergency fall back if something goes wrong
             self.blackboard.node.get_logger().error("Rank to ball had some issues")
-            raise ValueError("Count active Robots encountered to many robots")
+            raise ValueError(
+                f"Count active Robots encountered to many robots on the field: {number_of_active_teammates}"
```
jaagut:
```suggestion
                f"Count active Robots encountered too many robots on the field: {number_of_active_teammates}"
```

src/bitbots_navigation/bitbots_rust_nav/src/bin.rs
```diff
@@ -115,7 +115,10 @@ mod gui {
                 }
 
                 if let (Some(start), Some(goal)) = (self.start, self.goal) {
-                    let omap = ObstacleMap::new(self.config, self.obstacles.clone());
+                    let omap = ObstacleMap::new(
+                        self.config,
```
jaagut, line 119: So, this will only render round obstacles, correct?

jaagut (approved):
Is this safe or self-intersecting polygons? In general and when inflating them?

I guess, there is no reason to do this intentionally, but accidentally doing that could be bad.

Do we see the polygon in any visualization like in rviz?

src/bitbots_navigation/bitbots_rust_nav/src/bin.rs
Flova: Yes, the demo only uses round ones.

Flova:
> Is this safe or self-intersecting polygons? In general and when inflating them?

Not supported imo.

PR #918, opened 2026-06-12
Feature/deploy supervisor

### LINKED ISSUE #751: Improve field/deployment configuration mechanism
# Summary
We currently have field configurations for the blackboard.
We want to extend this configuration mechanism by the default field/ball/goal sizes for different leagues to support both the small, medium and large size of the humanoid league setup.

The field and ball to use should then be choosable during our deployment scripts.

jaagut:
## Summary

- Replace the sequential Fabric deployment pipeline with an interactive Textual supervisor.
- Reconcile each robot independently using shared OpenSSH ControlMaster connections.
- Add structured teamplayer status and control through a Unix socket.
- Support offline robot environments using lockfile-derived `pixi-pack` artifacts.
- Integrate fleet-wide match, WiFi, and ball configuration.
- Remove the legacy game-settings workflow.

## Motivation

The previous deployment tool synchronized robots at every pipeline step, used separate SSH connections for commands and rsync, produced noisy failure output, and made runtime configuration changes cumbersome.

The new supervisor maintains desired state per robot and continuously reconciles each robot independently. Operators can adjust configuration, inspect logs and progress, control teamplayer components, and attach to robot tmux sessions from one interface.

## Highlights

- Automatic robot discovery plus manually configured hosts
- Infinite reconnection attempts at a fixed configurable interval
- Independent deployment progress for each robot
- Desired-state generations with dependency-aware invalidation
- Concise error summaries with complete output available on demand
- Mouse and keyboard controls in the Textual TUI
- Shared SSH transport for commands, rsync, tmux, and forwarding
- Offline, resumable, checksum-verified environment provisioning
- Atomic runtime configuration and environment activation
- Fleet profiles with per-robot overrides
- `ball.diameter` published through `/parameter_blackboard`
- Teamplayer component status, logs, restart, start, and stop controls

## CLI

```text
pixi run deploy
pixi run deploy apply [targets and overrides]
pixi run deploy status [targets]
pixi run deploy cache prepare
```

## Testing

- 15 deployment supervisor tests passed
- 2 teamplayer Rust tests passed
- Selected ROS packages built successfully:
  - `bitbots_bringup`
  - `bitbots_ipm`
  - `bitbots_parameter_blackboard`
  - `bitbots_path_planning`
- Selected package tests passed
- Ruff, Cargo formatting, and `git diff --check` passed

## Known Validation Gap

No robot is currently available. Live SSH reconnection, LAN artifact transfer, NetworkManager activation, tmux attachment, and deployment to robot hardware remain to be validated separately.

## Dependency

This PR is stacked on #882 and should target `feature/teamplayer_tui`.

Closes #751


pixi.toml
```diff
@@ -191,6 +186,13 @@ clang-format = ">=21.1.0,<22"
 cppcheck = ">=2.18.3,<2.20"
 pre-commit = ">=4.4.0,<5"
 
+[feature.deploy.dependencies]
+gitpython = ">=3.1.45,<4"
+openssh = ">=10.2p1,<11"
+pixi-pack = ">=0.7,<1"
+rsync = ">=3.4.3,<4"
+textual = ">=8.2,<9"
```
Flova, line 194: I started the team player tui with textual, but had some performance problems. Especially when displaying large logs. Considering this build upon the team player tui it might be worth considering to do both in rust and have some proper API and or shared ui between them.

scripts/README.md
```diff
@@ -1,48 +1,111 @@
...
+There is one canonical robot environment: the normal Pixi `robot` environment
+at `.pixi/envs/robot`. Match operation, automated deployment, and interactive
+development all use this same environment.
+
+Robot networks do not have Internet access. While the deployment laptop can
+reach the package channels, prepare a package cache for the exact locked
+`linux-aarch64` robot environment:
+
+```shell
+pixi run deploy cache prepare
```
Flova, line 18: How is the situation on storage constraint environments? All deps are 10+ GB and we would need the 2-3 times in this case right?

scripts/README.md
```diff
@@ -1,48 +1,111 @@
...
+```
+
+`pixi-pack` creates a local Conda channel and collects the locked PyPI wheels in
+the user's XDG cache. Cache preparation limits downloads to four concurrent
```
Flova, line 22: I am not sure if we what to have easily outdated implementation information like the 4 concurrent workers in a readme. We should keep it condensed.

scripts/README.md
```diff
@@ -1,48 +1,111 @@
...
+machines.
+
+During deployment, the laptop temporarily serves this cache over the wired LAN.
+The supervisor gives `pixi install -e robot --frozen` a session-only mirror
+configuration, so Pixi installs or verifies the normal `.pixi/envs/robot`
+environment without contacting the Internet. Package hashes remain those from
+`pixi.lock`. The mirror configuration is passed only to the provisioning
+command and does not alter the robot's normal Pixi configuration.
+
+After deployment, automated commands use `pixi run -e robot --as-is` to avoid
```
Flova, line 33: We need to be really sure to not have code paths that don't sync and don't fail if the env is not up to date. Imo. an env that is not up to date should be a failure unless explicitly flagged by the user.

scripts/README.md
```diff
@@ -1,48 +1,111 @@
...
+implicit package operations. A developer connected by SSH can use Pixi normally
+without an activation wrapper:
+
+```shell
+pixi run -e robot build --packages-select bitbots_bringup
+pixi run -e robot test --packages-select bitbots_path_planning
```
Flova, line 39: Path planning is interesting, because it depends on the bitbots_cargo crate. Cargo behaves a bit like pixi and is a closed off system that is someehat similar to pixi. During build it downloads the necessary deps to statically link them. This would break the offline workflow. There might also be some additional cmake deps that fetch content during build, as this is pretty common for cmake projects. Cargo has a cargo fetch feature iirc. but this would need to be an additional explicit logic and I am not sure if I like the additional complexity it adds.

scripts/deploy/tests/test_artifacts.py
```diff
@@ -0,0 +1,139 @@
...
+import urllib.request
+from pathlib import Path
+from unittest.mock import patch
+
+import yaml
+from deploy.artifacts import ArtifactServer, ArtifactStore, EnvironmentArtifact
+
+
+def artifact_directory(root: Path) -> Path:
+    (root / "channel/linux-aarch64").mkdir(parents=True)
```
Flova, line 13: I think it would be nice to generally support deployment to x86 as well. Be it for testing or possible future robots.

src/bitbots_misc/bitbots_ipm/scripts/global_parameter_bridge.py
```diff
@@ -0,0 +1,90 @@
...
+from rclpy.node import Node
+from rclpy.parameter import Parameter, parameter_value_to_python
+
+
+class GlobalParameterBridge(Node):
+    def __init__(self) -> None:
+        super().__init__("ipm_global_parameter_bridge")
+        self._source = self.create_client(GetParameters, "/parameter_blackboard/get_parameters")
+        self._targets = {
+            "/soccer_ipm": (
```
Flova, line 14: I am not sure if the ball diameter is reconfigureable in all parts of the software.

src/bitbots_navigation/bitbots_path_planning/config/path_planning_parameters_template.yaml
```diff
@@ -35,7 +35,7 @@ bitbots_path_planning:
     ball_diameter:
       type: double
       default_value: 0.13
-      description: 'The diameter of the ball'
+      description: 'Fallback diameter of the ball; replaced from the global parameter blackboard at startup'
```
Flova, line 38: I am not in favor of these fallbacks. We should fail obviously if something went wrong

scripts/deploy/tests/test_artifacts.py
```diff
@@ -0,0 +1,139 @@
...
+        assert "solves = 1" in config
+    finally:
+        server.stop()
+
+
+def test_prepare_creates_lock_keyed_offline_channel(tmp_path: Path) -> None:
+    workspace = tmp_path / "workspace"
+    (workspace / "scripts/deploy").mkdir(parents=True)
+    (workspace / "scripts/deploy/pixi-pack-config.toml").write_text("[concurrency]\ndownloads = 4\n")
+    (workspace / "pixi.lock").write_text(
```
Flova, line 99: Why do we create our own channel?

scripts/deploy/models.py
```diff
@@ -0,0 +1,155 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any
+
+TEAMPLAYER_COMPONENTS = (
```
Flova, line 7: I don't know if I like the redundant definition here. Maybe we can do a yaml that both the team player tui and this reads?

scripts/deploy/models.py
```diff
@@ -0,0 +1,155 @@
...
+    "teamcom",
+    "world_model",
+    "whistle_detector",
+    "audio",
+    "tts",
+    "monitoring",
+    "record",
+)
+
+DEFAULT_COMPONENTS = frozenset(
```
Flova, line 25: Similarly the defaults are defined twice

scripts/deploy/tests/test_tui.py
```diff
@@ -0,0 +1,222 @@
...
+            assert app.query_one("#robot-nuc1", RobotCard).display
+            assert app.query_one("#robot-nuc2", RobotCard).display
+            assert str(app.query_one("#focus-toggle", Button).label) == "Focus"
+
+    asyncio.run(run())
+
+
+def test_space_uses_focused_widget_native_behavior() -> None:
+    profiles = ProfileStore.load(Path(__file__).parents[1])
+    supervisor = FakeSupervisor()
```
Flova, line 99: Ignore the specific line I just can't find the line anymore. At some point rm rf is used to clean the workspace. We should use pixi run clean for that, since there might be new artifacts to clean in the future.

scripts/README.md
jaagut: I see. That's great.

Alternatively, I thought of a method, that the deployment system could easily forward its own Internet connection. I didn't like it that much, as that would most definitely mean using a private hotspot, or team WiFi which is disallowed. Only another LAN connection could work, but there is never a guarantee for Internet in the vicinity of the field tables. Running a long cable from the team table to to the field is not viable either.

We can skip this completely of caching or forwarding if we can guarantee a fully functional env at the field by taking more time at the team table with all robots for preparation. At German open this was not possible...

Flova: I think for the world cup a more strict policy where we update the robots at the team table if deps are modified is the way to go. This is similar to the way it has been in the past when we used mainly apt dependencies. 

Forwarding the network would also be cool, but I agree with your points and additionally it seems to be highly dependent on the network setup of the host laptop.

PR #916, opened 2026-06-12
Smoothly transition between head modes using a spline

Flova:
## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot
- [ ] Create issues for future work
- [x] Triage this PR and label it

PR #919, opened 2026-06-12
Fix various bugs found during codebase audit

Flova:
- head_mover: fix sign error in path distance calculation (NaN risk), use pitch acceleration for pitch joint, fix double->int truncation of reduce_last_scanline, replace exact float comparison with epsilon
- odometry: initialize IMU mounting offset transform to identity, as the tf2 default constructor leaves it uninitialized
- world_model: fix inverted rotation in get_uv_from_xy and swapped sin/cos in get_xy_from_uv, remove dead falsiness guards on tuples
- pathfinding: fix angle normalization that produced negative rotation costs in time_to_ball, annotate get_ball_goal as Optional and handle None in go_to_ball
- misc: replace non-existent Duration.to_sec() with nanoseconds / 1e9
- animation: pass float timeout_sec to wait_for_server instead of a Duration object, which raised a TypeError when the server was not immediately available
- team_data: fix docstring of team_rank_to_ball
- costmap: normalize debug costmap to valid OccupancyGrid range [0, 100] instead of overflowing int8, guard against constant costmaps
- ball_filter: remove unused relative estimate lookup that silently skipped the negative observation update on tf errors
- team_communication: compare move_base_goal instead of a function against None, remove illusory hasattr checks on always-present proto3 fields, initialize timestamp state as Optional[TimeMsg]

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


ayin21:
Merge after #916

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/converter/message_to_team_data_converter.py
```diff
@@ -29,18 +29,16 @@ def convert(self, message: Proto.Message, team_data: TeamData) -> TeamData:
         return self.convert_optional_fields(message, team_data)
 
     def convert_optional_fields(self, message: Proto.Message, team_data: TeamData) -> TeamData:
-        if hasattr(message, "time_to_ball"):
-            team_data.time_to_position_at_ball = message.time_to_ball
+        # Note: these are non-optional proto3 scalar/enum fields, so they are always present
+        # (fields that were not set by the sender simply contain the default value)
+        team_data.time_to_position_at_ball = message.time_to_ball
```
texhnolyze, line 34: The reason we had these checks before was, that we actually extended the default protobuf protocol from https://github.com/RoboCup-Humanoid-TC/RobocupProtocol
with these additional fields.

The theory was that we could support interplay with other robots using the same protocol, even though this was never used.

texhnolyze: OK I actually tried this with a test case and since protobuf is backwards compatible this actually works as the fields are just filled with the default value (zero) as you commented.
Since our `role`/`action`/`offensive_side` are all enums with value 0 being `UNDEFINED` in the proto this works fine.
I've added my test case to this branch.
However, for `time_to_ball` an undefined value as defined as float in the proto will result in `0.0`, which would probably lead to issues as we would not go to the ball as we will think the other player is closer.

texhnolyze (approved):

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/converter/message_to_team_data_converter.py
Flova: So should we revert it or drop zero time to balls?

jaagut: Wouldn't we have the same situation regardless of reverting?

```py
message.time_to_ball  # Default 0.0
```

If I understood @texhnolyze correctly:

> for `time_to_ball` an undefined value as defined as float in the proto will result in `0.0`

```py
if hasattr(message, "time_to_ball"):  # Always exists
    message.time_to_ball  # Default 0.0
```

So this basically changes nothing, but adds a comment.

The actual question is, what happens with `TeamData` having `team_data.time_to_position_at_ball = 0.0` (or not setting it, as the default TeamData message content will also be 0.0?
This probably has not yet happened, as we never did proper inter-play.
I think, we should disregard `team_data.time_to_position_at_ball == 0.0` in every subscriber/consumer.

Flova:
This is weird. Why are your changes back here?

Flova:
@jaagut

jaagut:
Ohh. #928 merged into this branch instead of main. I guess, I assumed, this would be merged first...

jaagut:
I am starting to repair this now...

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/converter/message_to_team_data_converter.py
jaagut: 0799791 should address this

jaagut:
In team_communication_config.yaml, we still have a small value. Should we increase that?
```yaml
# average walking speed in [m/s]
avg_walking_speed: 0.2
```

Flova:
Yeah we can bump that to 0.7

src/bitbots_team_communication/bitbots_team_communication/test/converter/__snapshots__/test_state_to_message_converter.ambr
```diff
@@ -1,115 +0,0 @@
-# serializer version: 1
```
Flova, line 1: Why was this snapshot deleted?

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/converter/state_to_message_converter.py
```diff
@@ -158,18 +158,19 @@ def convert_time_to_ball(
...
-        if convert_target_position is not None:
+        if state.move_base_goal is not None:
             message = convert_target_position(state.move_base_goal, message)
         if state.ball is not None:
             message = convert_ball_position(state.ball, state.ball_velocity, state.ball_covariance, message)
         if state.seen_robots is not None:
             message = convert_seen_robots(state.seen_robots, message)
         if state.strategy is not None:
             message = convert_strategy(state.strategy, state.strategy_time, message)
-        if state.time_to_ball is not None:
```
Flova, line 169: Why is the none check removed while it is ceot for e.g. the strategy?

src/bitbots_team_communication/bitbots_team_communication/test/converter/__snapshots__/test_state_to_message_converter.ambr
jaagut: It has been replaced with values in the test itself. I believe it helped with the reusability of parts of the test.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/converter/state_to_message_converter.py
jaagut: I and apparently Codex believed, this `time_to_ball` would directly come from a message or proto. But that's not the case. Actually, state is the `TeamCommunication` class itself and the values are `None`-initialized. I will revert this and add better type hints, if possible.

ayin21: @jaagut ping

PR #922, opened 2026-06-18
Drop from WALKING to CONTROLLABLE when walk output has no significant motion

### LINKED ISSUE #881: Currently the walking instantly takes control of the robot. No animations other than the ones from the hcm can be played.
# Summary

## Steps to Reproduce

1.
2.
3.

Flova:
Tracks the last time any walk joint command deviated >3° from the previous
walk command. After standing_transition_delay seconds without such motion,
the HCM drops to CONTROLLABLE so animations can run while the walk node
continues balancing until the animation is requested. Significant motion (e.g. capture steps) immediately
returns the state to WALKING from CONTROLLABLE.

## Related issues

Closes #881

## Checklist

- [x] Run `pixi run build`
- [x] Write documentation
- [x] Test on your machine
- [x] Test on the robot

### DIFF
--- a/src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/decisions/walking.py
+++ b/src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/decisions/walking.py
@@ -4,12 +4,6 @@ from bitbots_hcm.hcm_dsd.decisions import AbstractHCMDecisionElement
 class RecentWalkingGoals(AbstractHCMDecisionElement):
     """
     Decides if the robot is currently walking or just balancing in place.
-
-    Returns STAY_WALKING when the walk node is active and any joint moved more
-    than 3° relative to the previous walk command within standing_transition_delay.
-    Returns NOT_WALKING (→ CONTROLLABLE) when the walk node stopped or has been
-    sending only small corrections for longer than standing_transition_delay.
-    The walk node continues to run and balance in CONTROLLABLE state.
     """
 
     def perform(self, reevaluate=False):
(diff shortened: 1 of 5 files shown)


src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/hcm_blackboard.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, node: Node):
 
         # Walking state
         self.last_walking_goal_time: Optional[Time] = None
+        # Time of the last walk motor goal that deviated >3° from the previous walk command
```
jaagut: in hcm cpp, you say its 0.5°... Probably this number should not be commented, instead the variable name

src/bitbots_motion/bitbots_hcm/bitbots_hcm/hcm_dsd/decisions/walking.py
```diff
@@ -3,26 +3,37 @@
 
 class RecentWalkingGoals(AbstractHCMDecisionElement):
     """
-    Decides if the robot is currently getting joint commands to from the walking node
+    Decides if the robot is currently walking or just balancing in place.
+
+    Returns STAY_WALKING when the walk node is active and any joint moved more
+    than 3° relative to the previous walk command within standing_transition_delay.
```
texhnolyze: What Jan said here as well, with a hard commented number

src/bitbots_motion/bitbots_hcm/src/hcm.cpp
```diff
@@ -155,7 +171,10 @@ class HCM_CPP : public rclcpp::Node {
   std::optional<sensor_msgs::msg::JointState> current_joint_state_;
 
   // Walking state
+  double significant_motion_threshold_ = 0.5 * M_PI / 180.0;  // default to 0.5 degrees in radians
```
texhnolyze, line 174: Should this be configurable?

Flova: It would be nice, but the hcmcpp does not load params from the normal hcm config file as it is a seperate node. I would have needed to add a full seperate config file for this one param, so I just opted for the hardcoded solution.

jaagut (approved):

PR #923, opened 2026-06-19
Spawn robot in sim in stable sit position like in real life

ayin21:
Also use same default kp/kd as on the real robot.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
+++ b/src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
@@ -88,6 +88,8 @@ class Simulation(Node):
             for idx in self._find_robot_indices()
         ]
 
+        self._apply_home_keyframe()
+
         self.time = 0.0
         self.time_message = Time(seconds=0, nanoseconds=0).to_msg()
         self.timestep = self.model.opt.timestep
@@ -108,6 +110,47 @@ class Simulation(Node):
             {"frequency": 4, "handler": lambda: self.publish(lambda robot: robot.publish_imu_event())},
             {"frequency": 32, "handler": lambda: self.publish(lambda robot: robot.publish_camera_event())},
         ]
+
+    def _apply_home_keyframe(self) -> None:
+        """Apply the 'home' keyframe joint values to all robots, leaving freejoints untouched.
+
+        The keyframe is defined in pi_plus.xml. Since attached sub-model keyframes are not
+        merged into the compiled world model, we load the robot model separately to read them.
+        """
+        robot_model_path = str(Path(self.package_path) / "xml" / "pi_plus.xml")
+        robot_model = mujoco.MjModel.from_xml_path(robot_model_path)
+        key_id = mujoco.mj_name2id(robot_model, mujoco.mjtObj.mjOBJ_KEY, "home")
+        if key_id < 0:
+            return
+
+        # Build a map from bare joint name → keyframe qpos value using the standalone robot model
+        home_qpos: dict[str, float] = {}
+        home_z: float | None = None
+        for i in range(robot_model.njnt):
+            jnt = robot_model.joint(i)
+            if jnt.type == mujoco.mjtJoint.mjJNT_FREE:
+                # qpos layout for freejoint: [x, y, z, qw, qx, qy, qz] — grab z (offset 2)
+                home_z = float(robot_model.key_qpos[key_id, jnt.qposadr[0] + 2])
+                continue
+            home_qpos[jnt.name] = robot_model.key_qpos[key_id, jnt.qposadr[0]]
+
+        for robot_sim in self.robots:
+            # Set z spawn height from keyframe while preserving x/y placement and orientation
+            if home_z is not None:
+                freejoint_name = f"robot_floating_base_joint_{robot_sim.robot.index}"
+                freejoint_id = mujoco.mj_name2id(self.model, mujoco.mjtObj.mjOBJ_JOINT, freejoint_name)
+                if freejoint_id >= 0:
+                    self.data.qpos[self.model.joint(freejoint_id).qposadr[0] + 2] = home_z
+
+            for joint in robot_sim.robot.joints:
+                # joint.ros_name is the bare name (e.g. "r_hip_pitch_joint") matching the robot model
+                value = home_qpos.get(joint.ros_name)
+                if value is None:
+                    continue
+                self.data.qpos[joint.joint_instance.qposadr[0]] = value
+                # For position actuators ctrl is the target position — same as the initial qpos
+                self.data.ctrl[joint.actuator_instance.id] = value
+        mujoco.mj_forward(self.model, self.data)
 
     def _generate_default_world(self) -> str:
         template_path = Path(self.package_path) / "xml" / "kid_field.xml"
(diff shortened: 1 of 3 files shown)


src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
```diff
@@ -109,6 +111,47 @@ def __init__(self):
             {"frequency": 32, "handler": lambda: self.publish(lambda robot: robot.publish_camera_event())},
         ]
 
+    def _apply_home_keyframe(self) -> None:
+        """Apply the 'home' keyframe joint values to all robots, leaving freejoints untouched.
+
+        The keyframe is defined in pi_plus.xml. Since attached sub-model keyframes are not
+        merged into the compiled world model, we load the robot model separately to read them.
+        """
+        robot_model_path = str(Path(self.package_path) / "xml" / "pi_plus.xml")
```
texhnolyze, line 120: I'm very confused what is going on here, without experience with the mujoco API.
It would be very nice to have more comments or even extract parts of this into named methods for better understandability.

Why can the same values for a joint in `home_qpos` be applied to both:
```
                self.data.qpos[joint.joint_instance.qposadr[0]] = value
                # For position actuators ctrl is the target position — same as the initial qpos
                self.data.ctrl[joint.actuator_instance.id] = value
```

Is this numpy and `joint.actuator_instance.id` is actually a list or something?



Also, would it not make sense to directly access `floating_base_joint` (our currently only `free` joint`), to make it clearer that's what we want so its easier to understand that this joint is only used to set the "spawn height", if that's even correct?

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
```diff
@@ -19,7 +19,7 @@
       <position kp="30" kv="0.6" forcerange="-10 10" ctrllimited="false"/>
     </default>
     <default class="pi_actuator_leg">
-      <position kp="30" kv="1.1" forcerange="-20 20" ctrllimited="false"/>
+      <position kp="100" kv="1.0" forcerange="-20 20" ctrllimited="false"/>
```
texhnolyze, line 22: Where do these changes come from and are they adapted in the `pi_plus.xml` in the mjlab and mujoco playgrounds?

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
```diff
@@ -109,6 +111,47 @@ def __init__(self):
...
+            if home_z is not None:
+                freejoint_name = f"robot_floating_base_joint_{robot_sim.robot.index}"
+                freejoint_id = mujoco.mj_name2id(self.model, mujoco.mjtObj.mjOBJ_JOINT, freejoint_name)
+                if freejoint_id >= 0:
+                    self.data.qpos[self.model.joint(freejoint_id).qposadr[0] + 2] = home_z
+
+            for joint in robot_sim.robot.joints:
+                # joint.ros_name is the bare name (e.g. "r_hip_pitch_joint") matching the robot model
+                value = home_qpos.get(joint.ros_name)
+                if value is None:
```
texhnolyze, line 148: Can it happen that a joint exists in the robot, but there is no value configured for in the keyframe?

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
Flova: The policies ship with their own pid params that are dynamically updated. These are for the default and mostly uses by legacy components like the animation server. The values are taken from the config used for the real robot, so this syncs the sim and robot config.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
Flova: Ideally not, but it could be imaginable that e.g. the head is not defined.

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
Flova: The mujoco API is like a c dev who was forced to write python, full of pointer like index lists etc..

The ctrl sets the setpoint for the actuator to hold the position and the qpos directly sets the simulator state to that position initially.

> Is this numpy and joint.actuator_instance.id is actually a list or something? 

What do you mean? we iterate through all joints and access the "numpy array" which is actually the exposed state of the mujoco C api. `joint.actuator_instance.id` is a scalar.

Flova: What do you mean with the last one?

jaagut:
@texhnolyze ping

src/bitbots_simulation/bitbots_mujoco_sim/bitbots_mujoco_sim/simulation.py
texhnolyze: I talked about this with @MegaIng. With the API I meant, that we have only a single free joint named `floating_base_joint`, which purpose as far as I understand it only is to be able to position the robot with it e.g. with a keyframe.

So I thought could we not just simplify the following code:
```python
# Build a map from bare joint name → keyframe qpos value using the standalone robot model
home_qpos: dict[str, float] = {}
home_z: float | None = None
for i in range(robot_model.njnt):
    jnt = robot_model.joint(i)
    if jnt.type == mujoco.mjtJoint.mjJNT_FREE:
        # qpos layout for freejoint: [x, y, z, qw, qx, qy, qz] — grab z (offset 2)
        home_z = float(robot_model.key_qpos[key_id, jnt.qposadr[0] + 2])
        continue

    home_qpos[jnt.name] = robot_model.key_qpos[key_id, jnt.qposadr[0]]

for robot_sim in self.robots:
    if home_z is not None:
        freejoint_name = f"robot_floating_base_joint_{robot_sim.robot.index}"
        freejoint_id = mujoco.mj_name2id(self.model, mujoco.mjtObj.mjOBJ_JOINT, freejoint_name)
        if freejoint_id >= 0:
            self.data.qpos[self.model.joint(freejoint_id).qposadr[0] + 2] = home_z
```

To something like:

```python
home_qpos: dict[str, float] = {}

for i in range(robot_model.njnt):
    jnt = robot_model.joint(i)
    home_qpos[jnt.name] = robot_model.key_qpos[key_id, jnt.qposadr[0]]

apply_z_spawn_height_from_keyframe(robot_model)

def apply_z_spawn_height_from_keyframe(robot_model):
  keyframe_floating_base_joint_id = ... Some way to get this
  keyframe_floating_base_joint = robot_model.joint(keyframe_floating_base_joint_id)

  model_floating_base_joint_name = f"robot_floating_base_joint_{robot_sim.robot.index}"
  model_floating_base_joint_id = mujoco.mj_name2id(self.model, mujoco.mjtObj.mjOBJ_JOINT, floating_base_joint_name)
  model_floating_base_joint = self.model.joint(floating_base_joint_id)

  # qpos layout for freejoint: [x, y, z, qw, qx, qy, qz] — grab z (offset 2)
  robot_spawn_heigth = float(robot_model.key_qpos[key_id, keyframe_floating_base_joint.qposadr[0] + 2])
  # Set z spawn height from keyframe while preserving x/y placement and orientation
  self.data.qpos[.qposadr[0] + 2] = robot_spawn_heigth
```

To make it clearer that is what is actually going on.

MegaIng (approved):
I understand the concerns raised by @texhnolyze , and agree in principle that a more explicit separation may look cleaner. However mujoco's API make all of them unwieldy - after multiple attempts I have not been able to find an alternative solution that actually looks better. I am going to merge the current version - if someone comes up with something noticeably better, please create a new PR.

PR #924, opened 2026-06-19
Multi robot sim fixes

ayin21:
A loose set of fixes that enables multi robot sim operations (very basic). For full team behavior testing a lot of things are still missing. For example the team communication does not work with multiple instances on the same machine.

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
```diff
@@ -78,32 +78,32 @@
       <camera name="track" pos="0.805 -1.105 0.2" xyaxes="0.814 0.580 0.000 -0.140 0.196 0.971" mode="trackcom"/>
 
       <inertial pos="-0.00041617 7.079e-05 -0.0392999" quat="0.707016 0.706374 -0.0255218 0.0226693" mass="1.4947" diaginertia="0.00182015 0.00167434 0.00111607"/>
-      <geom type="mesh" class="visual" rgba="0.752941 0.752941 0.752941 1" mesh="base_link"/>
+      <geom type="mesh" class="visual" rgba="0.1 0.1 0.1 1" mesh="base_link"/>
```
texhnolyze, line 81: Where do these changes come from and are they adapted in the `pi_plus.xml` in the mjlab and mujoco playgrounds?

src/bitbots_misc/bitbots_bringup/launch/mujoco_simulation.launch.py
```diff
@@ -81,6 +81,15 @@ def generate_domain_bridge_config(robot_domain: int, output_dir: Path) -> Path:
         "remap": f"{namespace}/joint_command",
     }
 
+    # Team communication: bridged bidirectionally on a single shared topic. Every robot's
+    # bridge mirrors this topic between its own domain and the shared main domain, merging
+    # all robots' traffic there and relaying it back out to everyone, including the sender
+    # (domain_bridge's bidirectional mode guards against the bridge looping on itself).
+    config["topics"]["team_comm_binary_transport"] = {
+        "type": "std_msgs/msg/UInt8MultiArray",
+        "bidirectional": True,
```
texhnolyze: Have you tested if two robots received each other's messages?

texhnolyze (commented):
This needs to be combined with:
- https://github.com/bit-bots/bitbots_main/pull/882 (changes in teamcomm threading)
- https://github.com/bit-bots/bitbots_main/pull/913 (changes in teamcomm UDP Broadcast address selection)

@Flova @jaagut in which order do we want to do this?
My suggestion would be to either merge this or the automatic broadcast address selection first and then the TUI PR first as it does not majorly have to do with teamcomm.

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
Flova: This just makes the robot a darker color to improve the robot and ball detection with multiple robots. The rl envs are not really affected by this.

src/bitbots_misc/bitbots_bringup/launch/mujoco_simulation.launch.py
Flova: I did a game with working team comm, but weirdly the current state here seems to be outdated. I need to check my PC if something was not pushed when I am back home.

Flova:
I would merge the TUI last as it has some TODOs still and is low prio. This here also has some todos left as for example currently all robots have the same player number, leading to the team com being dropped, so we need a non file based way to inject this. So the auto wifi one should go first imo..

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
texhnolyze: Ah yeah I didn't really have a look what the actual changes were :+1:

src/bitbots_misc/bitbots_bringup/launch/mujoco_simulation.launch.py
jaagut: Have you found something @Flova

Flova: I wasn't on the PC since then

MegaIng (changes requested):
I have an alterntive branch, I am going to open a PR later

PR #925, opened 2026-06-19
Better defence, goalie and supporter positions

ayin21:
TBD, small proof of concept:
 
image

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
How did you get this image?

Flova:
I vibe coded a simple GUI to interactively test the function that generates the placement.

This is the script I added in this pr. It is not ready yet.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+                "field.size.y",
+            ],
+        )
+
+        self._field = Field(
+            length=parameters["field.size.x"],
+            width=parameters["field.size.y"],
+            goal_width=parameters["field.goal.width"],
+        )
+        self._params = Params()
```
Flova, line 121: Ideally we want to load theses parameters from the behavior config file (or a separate one, as the current one is pretty large as it is).

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+        n_def += max(n - 4, 0)  # everyone past 4 is another defender
+
+        if has_support and n < 5 and ball is not None and field is not None:
+            own_third = -field.length / 2 + field.length / 3.0
+            if ball[0] < own_third:
+                has_support, n_def = False, n_def + 1
+
+        roles = []
+        if n >= 1:
+            roles.append("striker")
```
Flova: We might want to use a proper enum instead of strings

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+        pairs = [(old_poses[s_i], new_items[s_j][1], new_items[s_j][0])]
+
+        rem_i = [i for i in range(n) if i != s_i]
+        rem_j = [j for j in range(len(new_items)) if j != s_j]
+        if rem_i:
+            cost = np.empty((len(rem_i), len(rem_j)))
+            for a, i in enumerate(rem_i):
+                for b, j in enumerate(rem_j):
+                    op, npose = old_poses[i], new_items[j][1]
+                    cost[a, b] = np.linalg.norm(op[:2] - npose[:2]) + angle_w * self._angle_diff(op[2], npose[2])
```
Flova, line 404: We might want to use a slightly different cost calculation here. At the moment our cost calculation also includes the time to turn to the target position as well as the time to turn to the target orientation if we get close enough. But on the other hand this here would simplify the calculation significantly, so we maybe can just keep it.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+                bis = self._normalize(B - p) + self._normalize(opp - p)
+                head[role] = self._face(np.zeros(2), bis, fallback=self._face(p, opp))
+            else:
+                head[role] = self._face(p, B)  # goalie + defenders face the ball
+
+        return {role: np.array([p[0], p[1], head[role]]) for role, p in out.items()}
+
+    @cached_capsule_function
+    def get_formation_assignment(self) -> dict:
+        ballPose = self._blackboard.world_model.get_best_ball_point_stamped()
```
Flova: Maybe we should have separate functions for the team ball and "own ball" detection. Because we definitely want to use the team ball here.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -183,6 +184,26 @@ def get_active_teammate_poses(self, count_goalies: bool = False) -> list[Pose]:
                 poses.append(data.robot_position.pose)
         return poses
 
+    def quaternion_to_yaw(self, q) -> float:
```
Flova, line 203: Specify the quaternion notation xyzw vs wxyz and maybe move/check but it's utils for these functions.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -183,6 +184,26 @@ def get_active_teammate_poses(self, count_goalies: bool = False) -> list[Pose]:
...
+        for data in self.team_data.values():
+            if self.is_valid(data):
+                pose = data.robot_position.pose
+                robot_poses[data.robot_id] = [
+                    pose.position.x,
+                    pose.position.y,
+                    self.quaternion_to_yaw(pose.orientation),
+                ]
+        # include own data
+        robot_poses[self._blackboard.gamestate.get_own_id()] = list(self._blackboard.world_model.get_current_position())
```
Flova: Add a flag for that

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+        self._node.get_logger().info(f"Length of robot_poses: {len(robot_poses)}")
+
+        formation = self._compute_formation(ball, self._field, len(robot_poses), self._params)
+        new_items = list(formation.items())
+        ordered_jerseys = sorted(robot_poses.keys())
+        old_poses = [robot_poses[j] for j in ordered_jerseys]
+        pairs = self._match_assignment(old_poses, new_items, ball)
+
+        result = {}
+        for jersey, old_pose in zip(ordered_jerseys, [robot_poses[j] for j in ordered_jerseys]):
```
Flova: Reuse old poses var here

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,394 @@
...
+    @cached_capsule_function
+    def get_formation_assignment(self) -> dict:
+        ballPose = self._blackboard.world_model.get_best_ball_point_stamped()
+        ball = np.array([ballPose.point.x, ballPose.point.y])
+        robot_poses = self._blackboard.team_data.get_robot_poses()
+        self._node.get_logger().info(f"Length of robot_poses: {len(robot_poses)}")
+
+        formation = self._compute_formation(ball, self._field, len(robot_poses), self._params)
+        new_items = list(formation.items())
+        ordered_jerseys = sorted(robot_poses.keys())
```
Flova: I think we can simply this a lot by returning also indices or the player IDs from the matching.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_block_position.py
```diff
@@ -6,6 +6,26 @@
 from dynamic_stack_decider.abstract_action_element import AbstractActionElement
 from tf2_geometry_msgs import PoseStamped
 
+class GoToGoaliePosition(AbstractActionElement):
+    blackboard: BodyBlackboard
+    
+    def __init__(self, blackboard, dsd, parameters):
+        super().__init__(blackboard, dsd, parameters)
+
+    def perform(self, reevaluate=False):
```
Flova: Now it treads go to goalie, defense, etc all the same. That might be confusing, because somebody calls go to defense position to go explicitly there and he ends up at the goalie position.

Flova: I would prefer an explicit go to placement position. We can keep the old supporter position for the forced two robot ball touch, but that works outside of this placement. The PR is still pending, but it essentially removes the robot from the assignment after the ball was touched and goes to an explicit position to wait. While the others play with essentially one robot less. This can be done in the new concept, but we need the explicit support position for this. Other than that all other go to defense, block etc can be removed. Go to role needs to stay.

Flova: We also need to think about what to do if we don't see a ball. The position assignment should still work based on the lastest ball. I think the striker should perform our normal ball discovery behavior (turn around, go to opponent goal, go to middle line on the side). The rest of the robots should go to their role positions to not interfere with the striker during this procedure.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -33,34 +33,42 @@ def perform(self, reevaluate=False):
         self.publish_debug_data("time to ball", my_time_to_ball)
         self.publish_debug_data("Rank to ball", rank)
         if rank == 1:
-            return "YES"
+            return "NO"  # only for testing remeber to REMOVE!
```
Flova: Remove :P

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -33,34 +33,42 @@ def perform(self, reevaluate=False):
...
-            return "FIRST"
-        elif rank == 2:
-            return "SECOND"
-        elif rank == 3:
-            return "THIRD"
+        optimal_positioning = self.blackboard.positioning.get_formation_assignment()
+        own_position = optimal_positioning[self.blackboard.gamestate.get_own_id()]
+        role = own_position["role"]
+        self.publish_debug_data("Role from positioning", role)
+        if role == "striker":
```
Flova: Again we might want to use an enum here

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
Flova: I think we should remove the pass accept logic. That being said, the supporter should track the ball during the kick of the striker, we can keep that.

Flova, line 71: We need to be careful that the penalty shootout behavior still works if we use the striker positioning based on the new placement/ we need to find a good way to unify this.

Flova: Imo it should work out of the box as only one robot is present in this case but it needs testing.

Flova: This needs to be changed if the block position will be removed / handled by the new placement.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -94,10 +89,12 @@ $DoOnce
             NO --> $SecondaryStateTeamDecider
                 OUR --> #SearchBall
                 ELSE --> @AvoidBallActive, @LookAtFieldFeatures, @WalkInPlace + duration:2, @GoToRelativePosition + x:1 + y:0 + t:0, @Stand
```
Flova: I think we can do the search ball behavior in both cases here as our robots are way faster now and we had the issue in the past that we didn't leave the keep out area in time

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
Flova: Here is an issue I think because the block position is not really useful if we don't know where the ball is. We should handle the goalie like all other robots in this case.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -94,10 +89,12 @@ $DoOnce
...
-            YES --> $RankToBallNoGoalie
-                FIRST --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_first
-                SECOND --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_second
-                THIRD --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition
+            YES --> $RankToBallWithGoalie
+                STRIKER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_first
```
Flova: Don't go to defense position. Go to placement position or whatever we want to call it.

Flova: Also we have an issue here. We need an explicit placement for situations where the other team has e.g. a free kick. In this situation the striker role should be deactivated, the striker should be reassigned to a defensive role and we should enforce a keep out area around the ball. This should be pretty easily implemtable in the placement function if we add a flag for this.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -94,10 +89,12 @@ $DoOnce
             NO --> $SecondaryStateTeamDecider
                 OUR --> #SearchBall
                 ELSE --> @AvoidBallActive, @LookAtFieldFeatures, @WalkInPlace + duration:2, @GoToRelativePosition + x:1 + y:0 + t:0, @Stand
-            YES --> $RankToBallNoGoalie
-                FIRST --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_first
-                SECOND --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_second
-                THIRD --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition
+            YES --> $RankToBallWithGoalie
```
Flova: I don't think this selection is necessary anymore. Everyone should just go to their placement position and stand there if it was reached.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
Flova: Only the striker should search the ball. The others should go to the role position.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -110,16 +107,12 @@ $BallSeen
...
-            GOALIE --> #GoalieBehavior
-            ELSE --> $CountActiveRobotsWithoutGoalie
-                ONE --> $RankToBallNoGoalie
-                    FIRST --> #StrikerRole
-                    SECOND --> #DefensePositioning
-                ELSE --> $RankToBallNoGoalie
-                    FIRST --> #StrikerRole
-                    SECOND --> #SupporterRole
-                    THIRD --> #DefensePositioning
+        YES --> $RankToBallWithGoalie
```
Flova: I think the only really useful destinction here is striker, supporter, defender, goalie (maybe). We can probably do the same placement strategy for the goalie and and defenders and we are not limited in the number of defenders that way.

Flova: The striker needs special handling because it wants to kick if it reached its goal position. The supporter wants to track the ball if the striker is kicking and needs special handling for that and the others just go to their positions to stand in the way.

src/bitbots_team_communication/bitbots_team_data_sim_rqt/bitbots_team_data_sim_rqt/team_data_ui.py
```diff
@@ -45,6 +46,10 @@ def __init__(self, parent=None):
         self.main_layout.addWidget(self.state_box)
         self.time_to_position_box = TimeToPositionBox()
         self.main_layout.addWidget(self.time_to_position_box)
+        self.position_sliders_x = PositionCoordsX()
```
Flova, line 49: Add a toggle so some robot might not provide a ball position. Also add a slider for the covariance so the robot can tell how good it's ball is.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -110,16 +107,12 @@ $BallSeen
...
+            STRIKER --> #StrikerRole
```
Flova: The striker role needs to be reworked to position itself using the new striker placement.

Flova:
Also adapt the capsule/ gui so we can do headless tests of the placement function without the software stack. Currently the function is duplicated. 

Also probably some denouncing/hysteresis is needed. I would add it to the capsule as this needs to be global state. Just store if we have changed our role in the last x seconds and if did we want to stick with it. This prevents flickering behavior. The x can be randomly selected from a configurable range (0-1 sec for example).

src/bitbots_team_communication/bitbots_team_data_sim_rqt/bitbots_team_data_sim_rqt/team_data_ui.py
cleWu03: Yet the positions are the robots positions itself, not the ball position, but I will add that with as well as the suggested toggle and slider

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
cleWu03: i will take a closer look there once we have discussed how to handle set play situations with the new positioning capsule, got a view ideas there

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
cleWu03: I would use the positioning capsule with specialised formation functions for each set play situation, we should also integrate the "no double touch allowed" situation directly to the formation logic

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
cleWu03: I would use 1 = striker, 2 = goalie, 3, 5...x = defender 4 = supporter wich is the logical order for adding the roles depending on the number of robots. must be synchronised with capsule functions

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_block_position.py
cleWu03: the whole part with coding the positioning goals in the old GoTo decisions was only for testing :) I use GoToFormationPosition now in the most cases, we still need to discuss how to handle the supporter during no double touch, I would prefer a direct integration in the formation capsule

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
Flova: We can use a string enum.

src/bitbots_team_communication/bitbots_team_data_sim_rqt/bitbots_team_data_sim_rqt/team_data_ui.py
Flova: Ah nvm

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_ball.py
```diff
@@ -25,7 +26,15 @@ def __init__(self, blackboard, dsd, parameters):
         self.side_offset = parameters.get("side_offset", 0.00)
 
     def perform(self, reevaluate=False):
-        pose_msg = self.blackboard.pathfinding.get_ball_goal(self.target, self.distance, self.side_offset)
+        pose_msg = PoseStamped()
+        pose_msg.header.stamp = self.blackboard.node.get_clock().now().to_msg()
+        pose_msg.header.frame_id = self.blackboard.map_frame
+        optimal_positioning = self.blackboard.positioning.get_formation_assignment()
```
Flova: We should use the debounced assignment here.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_formation_position.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, blackboard, dsd, parameters):
+        super().__init__(blackboard, dsd, parameters)
+
+    def perform(self, reevaluate=False):
+        pose_msg = PoseStamped()
+        pose_msg.header.stamp = self.blackboard.node.get_clock().now().to_msg()
+        pose_msg.header.frame_id = self.blackboard.map_frame
+
+        optimal_positioning = self.blackboard.positioning.get_formation_assignment()
+        own_position = optimal_positioning[self.blackboard.gamestate.get_own_id()]
```
Flova: Same here

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_pass_position.py
Flova: We should remove these legacy roles

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -40,27 +40,34 @@ def get_reevaluate(self):
         return True
 
 
-class RankToBallNoGoalie(AbstractDecisionElement):
+class RankToBallWithGoalie(AbstractDecisionElement):
```
Flova: Don't call it rank to ball. Maybe something like assigned role.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -40,27 +40,34 @@ def get_reevaluate(self):
...
-        self.publish_debug_data("Rank to ball", rank)
-        if rank == 1:
-            return "FIRST"
-        elif rank == 2:
-            return "SECOND"
-        elif rank == 3:
-            return "THIRD"
+        role = self.blackboard.positioning.get_own_role()
+        self.publish_debug_data("Role from positioning", role)
+        return role
```
Flova: Writing this out even if there are many options helps the dsd parsing and language server.But it might be ugly so we need to consider what to do. Still it is currently not obvious what the possible states are.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -52,29 +52,11 @@ $GoalScoreRecently
...
-    NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @GoToBlockPosition
-
 #SupporterRole
-$PassStarted
-    YES --> $BallSeen
-        YES --> @TrackBall, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassAcceptPosition
-        NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassAcceptPosition
-    NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassPreparePosition
+$BallSeen
+    YES --> @TrackBall, @ChangeAction + action:positioning, @AvoidBallActive, @GoToFormationPosition
```
Flova: The supporter also needs localization, so we should not always track the ball. Just when a kick/pass is started by the striker. There is a decision for that.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
 $DoOnce
     NOT_DONE --> @ForgetBall, @LookAtFieldFeatures
     DONE --> $ConfigRole
```
Flova: This might be the initially designated goalie, which might be the striker now.

Flova: I would drop all of this. And consider it later in a more generic way.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
...
+        GOALIE --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @GoToFormationPosition
         ELSE --> $BallSeen
-            NO --> $SecondaryStateTeamDecider
-                OUR --> #SearchBall
-                ELSE --> @AvoidBallActive, @LookAtFieldFeatures, @WalkInPlace + duration:2, @GoToRelativePosition + x:1 + y:0 + t:0, @Stand
-            YES --> $RankToBallNoGoalie
-                FIRST --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_first
-                SECOND --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition + mode:freekick_second
-                THIRD --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition
+            NO --> #SearchBall
```
Flova: I would like a smarter search ball. All robots except for goaly first way a few seconds of the ball is lost to finish the head pattern. Then they turn for a few sec. If the ball was not found by then everyone returns to their config role position (not the dynamic one) except for the initial striker which performs the ball discovery behavior, which walks to the enemy goal etc.. The robot that was goalie before the ball was lost just stands still while the others turn up until the point where he goes the its initial role position again.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
...
+            YES --> $RankToBallWithGoalie
+                STRIKER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToFormationPosition //ugly yet
+                DEFENDER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToFormationPosition
+                SUPPORTER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition
+                GOALIE --> @GoToGoaliePosition
```
Flova: Why is this different?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_formation_position.py
```diff
@@ -0,0 +1,25 @@
...
+        pose = own_position["goal_pose"]
+        pose_msg.pose.position.x = pose[0]
+        pose_msg.pose.position.y = pose[1]
+        pose_msg.pose.orientation.w = pose[2]
+
+        self.blackboard.pathfinding.publish(pose_msg)
```
Flova: We should stop if we reached the target pose roughly. Make the threshold and whole feature a parameter, because e.g. the striker will not stop when he wants to strike while the supporter and goaly should. This is fairly important as walking on place for too long might break the localization due to slight drift that is unaccounted for in the linear noise model.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
...
 
 #Init
 @Stand + duration:0.1 + r:false, @ChangeAction + action:waiting, @LookForward, @Stand
 
 #NormalBehavior
 $SecondBallTouchAllowed
-    NO --> @SetNoSecondBallContactVariable + value:false + r:false, @LookAtFieldFeatures + r:false, @ChangeAction + action:passive + r:false, @AvoidBallActive + r:false, @GoToPassPreparePosition + blocking:true 
+    NO --> @SetNoSecondBallContactVariable + value:false + r:false, @LookAtFieldFeatures + r:false, @ChangeAction + action:passive + r:false, @AvoidBallActive + r:false, @GoToFormationPosition + blocking:true 
     YES --> $BallSeen
         NO --> $ConfigRole
```
Flova, line 89: Same issue with config role here.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
...
-                        FIRST --> #StrikerRole
-                        SECOND --> #DefensePositioning
-                    ELSE --> $RankToBallNoGoalie
-                        FIRST --> #StrikerRole
-                        SECOND --> #SupporterRole
-                        THIRD --> #DefensePositioning
-                        
+            YES --> $RankToBallWithGoalie
+                STRIKER --> #KickWithAvoidance
+                GOALIE --> @GoToFormationPosition
```
Flova: Why no team action and head behavior actions?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -91,40 +73,36 @@ $SecondaryStateTeamDecider
...
+                DEFENDER --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @GoToFormationPosition
```
Flova: Why not else?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -164,7 +142,7 @@ $IsPenalized
         SET --> $WhistleDetected
             DETECTED --> $SecondaryStateTeamDecider
                 OUR --> #PlayingBehavior
-                OTHER --> #PositioningReady
+                OTHER --> @GoToFormationPosition
```
Flova: Why not placing subtree? Are we allowed to go to the other teams side in this case? If so placing would be a good fit.

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -66,7 +66,7 @@ def __init__(self):
         self.run_spin_in_thread()
         self.try_to_establish_connection()
 
-        self.node.create_timer(1 / self.rate, self.send_message, callback_group=MutuallyExclusiveCallbackGroup())
+        self.node.create_timer(1 / self.rate, self.send_message, callback_group=ReentrantCallbackGroup())
```
Flova: This is dangerous! We might send two msgs at the same time. Why?

src/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication/bitbots_team_communication.py
```diff
@@ -264,9 +264,10 @@ def handle_message(self, string_message: bytes):
 
     def send_message(self):
         if not self.is_robot_allowed_to_send_message():
-            self.logger.debug("Robot is not allowed to send message")
+            self.logger.warm("Robot is not allowed to send message")
```
Flova: Throttle. Otherwise a penalized robot might spam warnings in the logs

Flova (changes requested):

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -187,6 +188,37 @@ def get_active_teammate_poses(self, count_goalies: bool = False) -> list[Pose]:
             if self.is_valid(data) and (data.strategy.role != Strategy.ROLE_GOALIE or count_goalies):
                 poses.append(data.robot_position.pose)
         return poses
+    
+    def get_index_of_passive_player(self) -> int:
```
Flova: There can be multiple players that send a passive msg (either by accident or intentionally) and we need to handle it. Because we designed passive to be relatively independent of the specific behavior I could imagine situations with multiple passive robots and I would expect it to still work if I set two robots to passive. On addition to that we might need explicit handling for our own robots passive flag. I don't know if it is included here.

Flova: Also the robots habe IDs. Index might be confusing and lead to people expecting it being continuous.

Flova (dismissed):

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_formation_position.py
cleWu03: right inside the action? how about a dsd decision either says stands or goes to formation position?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/actions/go_to_ball.py
cleWu03: what do you mean by that?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -72,36 +76,32 @@ $SecondaryStateTeamDecider
...
-        GOALIE --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @GoToFormationPosition
-        ELSE --> $BallSeen
-            NO --> #SearchBall
-            YES --> $RankToBallWithGoalie
-                STRIKER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToFormationPosition //ugly yet
-                DEFENDER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToFormationPosition
-                SUPPORTER --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToDefensePosition
-                GOALIE --> @GoToGoaliePosition
-                ELSE --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @AvoidBallActive, @GoToFormationPosition
+    DONE -->  $ReachedAndAlignedToPathPlanningGoalPosition + threshold:0.2 + latch:true
```
Flova: We need to send at least one goal with a do once before we can check this. To simiplify things we could also modify goToFormationPostion to send a stop command if we are close enough in position and orientation (controlled by a parameter)

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -72,36 +76,32 @@ $SecondaryStateTeamDecider
...
+        NO --> @ChangeAction + action:positioning, @LookAtFieldFeatures, @GoToFormationPosition
+
 
 #Init
 @Stand + duration:0.1 + r:false, @ChangeAction + action:waiting, @LookForward, @Stand
 
 #NormalBehavior
 $SecondBallTouchAllowed
-    NO --> @SetNoSecondBallContactVariable + value:false + r:false, @LookAtFieldFeatures + r:false, @ChangeAction + action:passive + r:false, @AvoidBallActive + r:false, @GoToFormationPosition + blocking:true 
+    NO --> $ReachedAndAlignedToPathPlanningGoalPosition + threshold:0.2 + latch:true
```
Flova: Same here

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -72,36 +76,32 @@ $SecondaryStateTeamDecider
...
             NO_NORMAL --> #StandAndLook
             NO_FREEKICK --> #Placing
-            YES --> $RankToBallWithGoalie
+            YES --> $AssignedRole
                 STRIKER --> #KickWithAvoidance
-                GOALIE --> @GoToFormationPosition
-                DEFENDER --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @GoToFormationPosition
                 SUPPORTER --> #SupporterRole
-                ELSE --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @GoToFormationPosition
+                ELSE --> $ReachedAndAlignedToPathPlanningGoalPosition + threshold:0.2 + latch:true
```
Flova: And here

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
```diff
@@ -150,8 +150,8 @@
             <geom name="head_pitch_link_collision0" class="collision" pos="0.026 0 0.055" type="box" size="0.05 0.055 0.0225"/>
             <body name="camera_link" pos="0.048925 0.0053519 0.058472">
               <inertial pos="0.00963743 0.00075253 0.00058537" quat="0.70738 0.706786 0.00245222 0.00788061" mass="0.0584" diaginertia="7.26043e-05 7.249e-05 5.51563e-06"/>
+              <camera name="head_camera" focalpixel="362.6533508300781 362.6533508300781" resolution="672 376" sensorsize="672 376" pos="0.03 0.024 0.0" xyaxes="0.0 -1.0 0.0 0.5 0.0 0.866"/>
```
Flova, line 153: What changed with the cam?

jaagut:
@copilot, behebe bitte die Mergekonflikte in diesem Pull Request.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
MegaIng: Surely we already have a function for this *somewhere*. We should not be reimplementing the logic here.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -188,6 +189,39 @@ def get_active_teammate_poses(self, count_goalies: bool = False) -> list[Pose]:
...
+        if self.strategy.action == Strategy.ACTION_PASSIVE:
+            passive_robot_ids.append(self._blackboard.gamestate.get_own_id())
+        return passive_robot_ids
+
+    def quaternion_to_yaw(self, q) -> float:
+        """Extract yaw (theta) from a quaternion."""
+        return math.atan2(2.0 * (q.w * q.z + q.x * q.y), 1.0 - 2.0 * (q.y * q.y + q.z * q.z))
+
+    def get_robot_poses(self, include_own: bool = True) -> dict[int, list[float]]:
+        """Returns a mapping of jersey_number -> [x, y, theta] for all active robots."""
```
MegaIng, line 208: We should be using consistent terminology, i.e. robot id instead of jersey_number.

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/decisions/closest_to_ball.py
```diff
@@ -40,27 +40,16 @@ def get_reevaluate(self):
...
-            return "SECOND"
-        elif rank == 3:
-            return "THIRD"
-        else:
-            # emergency fall back if something goes wrong
-            self.blackboard.node.get_logger().warning("Rank to ball had some issues")
-            return "FIRST"
+        role = self.blackboard.positioning.get_own_role()
+        self.publish_debug_data("Role from positioning", role)
+        return role.upper()
```
MegaIng, line 52: DSD is currently designed to assumed a fixed, known list of return values. But the new role names are partially auto generated. Do we want to
- ignore this? (hoping we only have one class of auto generated names which can be covered by ELSE)
- Update DSD to support pattern matching or something?
- Removing the extra index and unified all defenders into a single role name for the purpose of the DSD?

src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -60,31 +60,13 @@ $GoalScoreRecently
...
-$PassStarted
-    YES --> $BallSeen
-        YES --> @TrackBall, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassAcceptPosition
-        NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassAcceptPosition
-    NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToPassPreparePosition
+$BallSeen
+    YES --> $PassStarted
+        YES --> @TrackBall, @ChangeAction + action:positioning, @AvoidBallActive, @GoToFormationPosition
+        NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToFormationPosition + stand:true + enter_position:%support_enter_position + leave_position:%support_leave_position + enter_orientation:%support_enter_orientation + leave_orientation:%support_leave_orientation
+    NO --> @LookAtFieldFeatures, @ChangeAction + action:positioning, @AvoidBallActive, @GoToFormationPosition + stand:true + enter_position:%support_enter_position + leave_position:%support_leave_position + enter_orientation:%support_enter_orientation + leave_orientation:%support_leave_orientation
```
MegaIng, line 68: I think it would make sense to do something so that these parameter names don't need to be repeated - i.e. `GoToFormationPosition` gets the name of the role (or looks it up), and then auto selects the relevant parameters.

MegaIng (changes requested):
I have not reviewed the big files: positioning_capsule, go_to_formation_position, team_data_ui.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+            if clear_radius is not None and ball is not None:
+                self._clear_ball(positions, ball, clear_radius, field)
+            disp = {n: np.zeros(2) for n in names}
+            for i in range(len(names)):
+                for j in range(i + 1, len(names)):
+                    a, b = names[i], names[j]
+                    diff = positions[a] - positions[b]
+                    dist = np.linalg.norm(diff)
+                    if dist < sep:
+                        dirv = self._normalize(diff, np.array([0.0, 1.0])) if dist > 1e-6 else np.array([0.0, 1.0])
```
MegaIng: `self._normalize` already has a fallback mechanisim. Duplicating it here is a waste of time and reduces maintainability.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+        opp = np.array([+field.length / 2.0, 0.0])
+
+        d = float(np.linalg.norm(b - goal))
+        to_ball = self._normalize(b - goal)  # our-goal -> ball
+        perp = np.array([-to_ball[1], to_ball[0]])
+
+        roles = self._allocate_roles(n_players, b, field, include_supporter=params.include_supporter)
+        if opp_set_play and Role.SUPPORTER in roles:
+            n_def = sum(1 for r in roles if r.startswith(Role.DEFENDER + "_"))
+            roles[roles.index(Role.SUPPORTER)] = Role(f"{Role.DEFENDER}_{n_def}")
```
MegaIng: Instead of awkwardly removing a defender afterwards, instead the include_supporter parameter should be made dependent on `opp_set_play`.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+                h = max(field.goal_width / 2 - params.post_margin, 0.0)  # safe half-mouth
+                # aim at the goal, target clamped into the safe mouth: this gives a straight
+                # (+x) shot whenever the ball is aligned within the posts, angled otherwise
+                target = np.array([opp[0], np.clip(b[1], -h, h)])
+                aim_goal = self._normalize(target - b)
+                # fallback: play back toward our side (field centre)
+                aim_back = self._normalize(np.array([0.0, 0.0]) - b, fallback=np.array([-1.0, 0.0]))
+                # blend to back-pass only when close to the opp goal AND not aligned
+                near_goal = self._smoothstep(b[0], opp[0] - params.back_dist - 1.0, opp[0] - params.back_dist)
+                not_aligned = self._smoothstep(abs(b[1]), h, h + 1.0)
```
MegaIng, line 477: The `- 1.0` and `+ 1.0` are both magic values that should be parameters. I also feel like that the current parameters `back_dist` and `post_margin` are significantly falsified in their interpretability by the way they are currently used.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+                w_back = near_goal * not_aligned
+                # rotate the aim around the ball at constant angular rate (shortest path) so
+                # the striker swings smoothly rather than whipping when the two aims oppose
+                a0 = np.arctan2(aim_goal[1], aim_goal[0])
+                a1 = np.arctan2(aim_back[1], aim_back[0])
+                da = (a1 - a0 + np.pi) % (2 * np.pi) - np.pi  # shortest signed turn
```
MegaIng: Here the `self._angle_diff` helper should be used.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+                out[Role.STRIKER] = b - params.kick_offset * aim
+                head[Role.STRIKER] = ang  # striker faces where it kicks
+                kick_aim = aim
+
+        # --- goalie: on the ball->goal axis, hugging the goal, clamped to the mouth -- #
+        if Role.GOALIE in roles:
+            g = goal + params.d_g * to_ball
+            g = np.array(
+                [
+                    np.clip(g[0], -field.length / 2 + 0.05, -field.length / 2 + params.d_g),
```
MegaIng, line 495: Where does `0.05` come from? Should probably also be a parameter.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+    A short pairwise-repulsion pass at the end enforces a minimum separation; it is
+    also what shoves the defenders sideways into clean flanking slots when they would
+    otherwise pile onto the goalie -> the "wall next to the goalie" emerges for free.
+    """
+
+    # goalie
+    d_g: float = 0.55  # how far the goalie comes out of the goal (dist from goal centre)
+    # defenders
+    alpha: float = 0.42  # defender depth as a fraction of |ball-goal|  (push-up factor)
+    depth_bias: float = 0.0  # extra defender depth: + = further forward, - = further back
```
MegaIng, line 81: These parameters could be better described IMO. I originally interpreted `depth_bias` as making the defenders form into some form of half-circle. But it actually moves around the entire defender line.

`alpha` also isn't a good name.

In generally, we don't need to save letters for these names, it's more important that they are self describing (which most of them are).

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+        # --- orientations (computed from the final positions) --------------------- #
+        for role, p in out.items():
+            if role in head:
+                continue  # striker already set (faces its kick aim)
+            if role == Role.SUPPORTER:
+                # face the bisector of (toward ball, toward opp goal): "watch both"
+                bis = self._normalize(b - p) + self._normalize(opp - p)
+                head[role] = self._face(np.zeros(2), bis, fallback=self._face(p, opp))
+            else:
+                head[role] = self._face(p, b)  # goalie + defenders face the ball
```
MegaIng: Incorrect comments are worse than no comments: This case can actually also be used by the striker in case of `opp_set_play`.

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,539 @@
...
+        """C1 ramp from 0 (x<=lo) to 1 (x>=hi)."""
+        if hi <= lo:
+            return float(x >= hi)
+        t = np.clip((x - lo) / (hi - lo), 0.0, 1.0)
+        return float(t * t * (3 - 2 * t))
+
+    @staticmethod
+    def _angle_diff(a: float, b: float) -> float:
+        """Smallest absolute angle between two headings (rad), in [0, pi]."""
+        return float(abs((a - b + np.pi) % (2 * np.pi) - np.pi))
```
MegaIng: Let's use `math.remainder` (not `np.remainder`) instead of the slightly confusing math.

MegaIng (changes requested):

MegaIng:
Remaining issues

- move parameters, add more parameters (various, check PR comments once someone tackles this)
- proper handling of own ball vs team ball, unified team ball filter
- check penalty shootout behavior
- smarter ball search (@flova, your suggestion may have already been implemented, then this is obsolete, I wasn't sure)
- Rethink DSD in the presences of dynamically generated dispatch roles.

I think these can all be opened as follow up issues.

Otherwise I think this is ready to merge. I would like a glancing review from @Flova and a somewhat careful review from @jaagut, ideally tomorrow so that this is ready for IFA.

MegaIng (approved):

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/team_data_capsule.py
```diff
@@ -188,6 +189,35 @@ def get_active_teammate_poses(self, count_goalies: bool = False) -> list[Pose]:
...
+        data: TeamData
+        for data in self.team_data.values():
+            if self.is_valid(data) and (data.strategy.action == Strategy.ACTION_PASSIVE):
+                passive_robot_ids.append(data.robot_id)
+        if self.strategy.action == Strategy.ACTION_PASSIVE:
+            passive_robot_ids.append(self._blackboard.gamestate.get_own_id())
+        return passive_robot_ids
+
+    def get_robot_poses(self, include_own: bool = True) -> dict[int, list[float]]:
+        """Returns a mapping of robot id -> [x, y, theta] for all active robots."""
```
jaagut: This claims, it only returns from the active robots, but never checkes whether they are actually active or passive.

src/bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml
```diff
@@ -45,8 +46,27 @@ body_behavior:
           1: [ -0.085, 0.33 ]
           2: [ -0.085, -0.33 ]
 
+    # Whether the formation assignment includes a supporter role.
+    # If false, the supporter's slot is filled by an extra defender instead.
+    formation_include_supporter: false
+
+    support_enter_position: 0.1
```
jaagut, line 54:
```suggestion
    # Hysteresis for being considered at role pose
    support_enter_position: 0.1
```

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,544 @@
...
+    width: float = 6.0  # y extent
+    goal_width: float = 1.5  # goal mouth
+    margin: float = 0.3  # keep field players this far inside the touchlines
+
+
+@dataclass
+class Params:
+    """
+    A single pure function `compute_formation(ball, field, n_players, params)` maps a
+    ball position to positions for every non-... well, *every* role (goalie, defenders,
```
jaagut: ...

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
jaagut: We should consider making all/most of the parameters and calculations relative to the field size i.e., normalize them.

src/bitbots_behavior/bitbots_body_behavior/config/body_behavior.yaml
```diff
@@ -24,17 +25,17 @@ body_behavior:
     # To be useful for different field sizes, use values in [-1, 1] for x and y
 
     role_positions:
-      goalie: [ -0.7, 0.0 ]
+      goalie: [ -0.9, 0.0 ]
```
jaagut, line 28: We deleted surprisingly few parameters. Please recheck.

jaagut (commented):
wip

src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/positioning_capsule.py
```diff
@@ -0,0 +1,544 @@
...
+        """Return the role of the robot with the own_id."""
+        own_id = self._blackboard.gamestate.get_own_id()
+        assignment = self.get_formation_assignment()
+        return Role(assignment[own_id]["role"])
+
+    def _get_formation_assignment(self, set_play=False) -> dict[int, RobotAssignment]:
+        ball_pose = self._blackboard.world_model.get_best_ball_point_stamped()
+        ball = np.array([ball_pose.point.x, ball_pose.point.y])
+        robot_poses = self._blackboard.team_data.get_robot_poses()
+        passive_robot_ids = self._blackboard.team_data.get_id_of_passive_player()
```
jaagut, line 157:
```suggestion
        passive_robot_ids = self._blackboard.team_data.get_ids_of_passive_players()
```
This should get renamed

jaagut (approved):

Issue #258, opened 2023-02-27
Migrate `speak_ip` to new tts


jaagut:
## Is your feature request related to a problem? Please describe.
With bit-bots/humanoid_league_misc/pull/130, a new and better sounding tts (Text to speech) was introduced.

## Describe the solution you'd like
Please migrate the `speak_ip.py` script to the new solution.
This enables us to drop the `pyttsx3` dependency:

https://github.com/bit-bots/bitbots_meta/blob/546b375914ab6d3bdbe1c7c5ce027f0149f11dde/requirements.txt#L7

MegaIng:
Done in ansible. We are not using the new TTS, but espeak because of the inability to speak ip addresses.

PR #929, opened 2026-06-21
Add speak_ip.py

MegaIng:
# Summary

For use in a systemd service, this script adds the ability to automatically detect "relevant" IP addresses and speak them out loud.

Uses espeak instead of our normal tts by default because the default TTS results in sometimes incomprehensible outputs.

Uses minimal external dependencies, especially not ros, even when using the diffusion model,

## Related issues

#258

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/src/bitbots_misc/bitbots_tts/scripts/speak_ip.py
+++ b/src/bitbots_misc/bitbots_tts/scripts/speak_ip.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""Script to speak out all relevant IP addresses and their networks."""
+
+import argparse
+import os
+
+
+def get_relevant_ip_addresses() -> list[tuple[str, str]]:
+    """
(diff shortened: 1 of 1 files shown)


Flova:
Maybe this would be better suited to live in the ansible repo imo.. Especially when it's not dependent on other Bit-Bots Software.

MegaIng:
I was considering that it may be useful here incase we ever want to implement a way to trigger it via some button or action.

And it can technically depend on the bitbots tts stuff, that's just disabled by default, incase we ever figure out properly what is wrong and how to fix it.

Flova:
I agree that in these cases it would make sense to keep it here. But we can also easily move it here again if we implement these features.

Flova:
Moved to ansible

PR #930, opened 2026-06-21
Rl getup policy

Flova:
# Summary
I need to go home now but works on the real robot

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot


Flova:
I did some further tuning. Maybe needs testing on the robot again.

ayin21 (approved):

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/joint_handler.py
```diff
@@ -71,14 +71,27 @@ def get_velocity_data(self) -> np.ndarray:
         )
         return joint_velocities
 
-    def get_joint_commands(self, onnx_pred) -> JointCommand:
+    def get_joint_commands(self, onnx_pred, relative_to_current: bool = False) -> JointCommand:
```
jaagut, line 74: I would like to have a doc string here

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/nodes/mjlab_getup_node.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # We got dangerously close to a self collision in the hip and to be safe we clip the pitch angles.
+        # The policy is able to work around this limitation
+        def clipping(name, pos):
+            if "hip_pitch" in name:
+                return np.clip(pos, -1.5, 1.5)
+            else:
+                return pos
+
+        joint_command.positions = [
```
jaagut, line 63: Couldn't we do this with indexing, as I presume the arrays are in constant order? That would avoid a lot of string comparisons.

Flova: Probably, but we dont want this to ever be applied to the wrong joint. I am fine with a few hundred string comparisons per tick for this.

PR #933, opened 2026-06-24
Feature/deploy multiple ips per robot main

jaagut:
# Summary

This has already been approved in #928, but I did a git oopsy and that landed on the wrong branch. Please re-review this.

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


.github/workflows/ci.yml
```diff
@@ -45,3 +45,6 @@ jobs:
 
       - name: Test
         run: pixi run -e default test
+
+      - name: Test deploy scripts
```
Flova, line 49: I am not too happy to have a second code path for the deployment tests. Can we just make the deployment a package at some point and run the tests that way?

jaagut: I think this will not live here long-term. With the new TUI, I can build that as a package.

Flova (approved):
But make an issue to move it to its own pkg with its own tests

PR #936, opened 2026-06-29
Feature/ht kick

jgueldenstein:
# Summary
Include kick policy from hightorque in our rl framework

called via an action, incorporation in behavior pending

the policy could also do walking but I propose to walk to a suitable position with our own policy and activate the kick policy from there.

preliminary tests on robot worked well, kick direction is not so accurate, kick strength matters quite a lot.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] test in simulation
- [x] Test on the robot

TODO:
- [x] switch between policies when action is called
- [x] cleanup


jgueldenstein:
closes #937

jgueldenstein:
tested on the robot and working well, does not do anything as long as the action is not called.
transition out of kick is still missing, the policy should run with a zero comand for a set time after the kick, but feel free to review already

src/bitbots_misc/bitbots_bringup/launch/simulator_teamplayer.launch
```diff
@@ -1,20 +1,23 @@
...
+    <arg name="localization" default="false" description="Whether the localization system should be started" />
     <arg name="motion" default="true" description="Whether the motion control system should be started" />
-    <arg name="path_planning" default="true" description="Whether the path planning should be started"/>
+    <arg name="path_planning" default="false" description="Whether the path planning should be started"/>
     <arg name="teamcom" default="false" description="Whether the team communication system should be started" />
     <arg name="vision" default="true" description="Whether the vision system should be started" />
     <arg name="world_model" default="true" description="Whether the world model should be started"/>
-    <arg name="rl_motion" default="true" description="Whether to use the RL motion system instead of the regular one" />
+    <arg name="rl_motion" default="false" description="Whether to use the RL motion system instead of the regular one" />
+    <arg name="web" default="true" description="Use web-based mjviser viewer instead of the native MuJoCo viewer"/>
```
jaagut: Do you want to merge these changes?

src/bitbots_motion/bitbots_hcm/src/hcm.cpp
```diff
@@ -117,6 +119,20 @@ class HCM_CPP : public rclcpp::Node {
...
 
+  void kick_goal_callback(bitbots_msgs::msg::JointCommand msg) {
+    // The kick is sending us joint goals; remember when the last one arrived so
+    // the DSD (RecentKickGoals) can tell whether a kick is currently active.
+    last_kick_goal_time_ = msg.header.stamp;
+
+    // Forward the kick goal to the motors if we are in a state that allows it.
+    // CONTROLLABLE lets the first goal through (which then transitions us to
+    // KICKING), and KICKING keeps the following goals flowing.
+    if (current_state_ == bitbots_msgs::msg::RobotControlState::CONTROLLABLE ||
```
Flova, line 133: This is a bit risky as the walking also talks during controllable. The state essentially is "walking ready to give up control". And the argument that we would drop the first tick otherwise also only applies of we kick if we are in controllable (e
g. standing stable for one second or more). In all other situations the first tick is indeed dropped. The line here could enable a situation where both talk during controllable which might be pretty bad.

src/bitbots_misc/bitbots_bringup/launch/simulator_teamplayer.launch
```diff
@@ -1,20 +1,23 @@
 <?xml version="1.0"?>
 <launch>
     <arg name="audio" default="false" description="Whether the audio system should be started" />
-    <arg name="behavior" default="true" description="Whether the behavior control system should be started" />
+    <arg name="behavior" default="false" description="Whether the behavior control system should be started" />
```
Flova: Revert these

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
```diff
@@ -1,33 +1,50 @@
...
+import tf2_ros
+from geometry_msgs.msg import PoseStamped, PoseWithCovarianceStamped
+from rclpy.duration import Duration
+from rclpy.time import Time
 
 from handlers.handler import Handler
 
+try:
+    from bitbots_tf_buffer import Buffer, TransformListener
+except ImportError:
```
Flova: At least print a warning in this case.

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
```diff
@@ -1,33 +1,50 @@
...
+
+    def get_ball_pos(self) -> np.ndarray:
+        assert self._ball_pose is not None
+        pose_stamped = PoseStamped()
+        # Use Time(0) to get the latest available transform regardless of message age
+        pose_stamped.header.frame_id = self._ball_pose.header.frame_id
+        pose_stamped.header.stamp = Time(seconds=0).to_msg()
+        pose_stamped.pose = self._ball_pose.pose.pose
+        try:
+            transformed = self._tf_buffer.transform(pose_stamped, "base_footprint", timeout=Duration(seconds=0.1))
```
Flova: Did we try base_link?

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
```diff
@@ -1,33 +1,50 @@
...
+            return np.array([transformed.pose.position.x, transformed.pose.position.y], dtype=np.float32)
+        except (tf2_ros.LookupException, tf2_ros.ConnectivityException, tf2_ros.ExtrapolationException) as e:
+            self._node.get_logger().warn(f"TF transform to base_footprint failed: {e}", throttle_duration_sec=1.0)
+            pos = self._ball_pose.pose.pose.position
```
Flova: I don't know if we want to do this. This is always the wrong move. I would just use the last one or a bounded default like 1.0, 0.0.

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+        self._base_footprint_frame = str(node.declare_parameter("base_footprint_frame", "base_footprint").value)
+        # tf buffer to read the robot's yaw in odom (odom -> base_footprint).
+        self._tf_buffer = Buffer(node=self._node)
+
+        self._cmd_vel: Optional[Twist] = None
+        self._ball_pos: Optional[np.ndarray] = None
+
+        # kick state: set by the action (its own thread), read by the control loop
+        self._kick_active = False
+        self._post_kick_stand = False
```
Flova, line 84: If we keep this I would prefer a state enum that makes impossible states impossible to represent. For the line and the one above

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+        # The kick action runs in its own (reentrant) callback group so its
+        # timed execute callback can run concurrently with the control loop.
+        self._kick_action_server = ActionServer(
+            self._node,
+            Kick,
+            "rl_kick",
+            execute_callback=self._execute_kick,
+            goal_callback=self._goal_callback,
+            cancel_callback=self._cancel_callback,
+            callback_group=ReentrantCallbackGroup(),
```
Flova, line 108: Mutable exclusive cb is sufficient here

src/bitbots_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+        # history state (newest at index -1), lazily initialized on first update
+        self._clean_hist: Optional[np.ndarray] = None  # (H, 7), pushed every step
+        self._soccer_hist: Optional[np.ndarray] = None  # (H, 7), pushed every pub_period
+        self._pub_counter = 0
+
+    # ------------------------------------------------------------------ #
+    # subscriptions
+    # ------------------------------------------------------------------ #
+
+    def _ball_callback(self, msg: BallArray) -> None:
```
Flova: Who calls this?

Flova: Duplicate with ball handler?

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+    # subscriptions
+    # ------------------------------------------------------------------ #
+
+    def _ball_callback(self, msg: BallArray) -> None:
+        if msg.balls:
+            center = msg.balls[0].center
+            self._ball_pos = np.array([center.x, center.y], dtype=np.float32)
+
+    def has_data(self) -> bool:
+        # Non-blocking: missing cmd_vel/ball just yield zeros, so the policy can
```
Flova, line 138: But should it run in this case?

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+                Time(),
+                Duration(seconds=timeout_s),
+            )
+        except Exception as e:  # tf2 raises various Lookup/Extrapolation errors
+            self._node.get_logger().warning(f"Could not look up robot yaw in odom: {e}", throttle_duration_sec=1.0)
+            return None
+        q = tf.transform.rotation
+        return _yaw_from_quat_xyzw(q.x, q.y, q.z, q.w)
+
+    def _compute_kick_dir_b(self) -> float:
```
Flova, line 173: Better naming

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+        """Kick heading expressed in the current base frame.
+
+        Re-projects the odom-anchored heading into the current base frame each
+        step. Falls back to the body-frame heading recorded at receipt if the
+        kick was never anchored, and holds the last good value if a per-step
+        lookup transiently fails.
+        """
+        if self._kick_dir_odom is None:
```
Flova, line 181: Ar least print a warning. I don't like implicit fallbacks as they make things really tedious to debug later. Things should fail fast and verbose if they fail.

src/bitbots_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # Anchor the body-frame heading into odom so it stays world-fixed for the
+        # duration of the kick. If tf is unavailable, fall back to a body-fixed
+        # heading (kick_dir_odom = None). Set the anchor BEFORE marking the kick
+        # active so the control loop never reads a half-initialized anchor.
+        anchor_yaw = self._robot_yaw_odom(timeout_s=0.1)
+        if anchor_yaw is not None:
+            self._kick_dir_odom = _wrap_to_pi(anchor_yaw + self._kick_dir_body)
+        else:
+            self._kick_dir_odom = None
```
Flova: What happens in that case?

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
```diff
@@ -0,0 +1,315 @@
...
+            f"Kick started: dir={math.degrees(self._kick_dir_body):.1f} deg (body), "
+            f"strength={self._kick_speed:.2f}, timeout={timeout:.2f} s"
+        )
+
+        end = self._node.get_clock().now() + Duration(seconds=timeout)
+        feedback = Kick.Feedback()
+        result = Kick.Result()
+
+        while self._node.get_clock().now() < end:
+            if goal_handle.is_cancel_requested:
```
Flova, line 246: We might want to cancel ourselves if we are in the wrong robot state. Especially with the fast standup it could happen that we fall and standup before the kick has finished leading to weird behavior. It would be easier to self cancel in that case.

src/bitbots_motion/bitbots_rl_motion/launch/rl_motion.launch
```diff
@@ -2,6 +2,7 @@
   <arg name="sim" default="false"/>
   <arg name="walk" default="true"/>
   <arg name="kick" default="false"/>
+  <arg name="kick_ball" default="true"/>
```
Flova: Why not reuse the kick flag?

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/nodes/kick_ball_node.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def main():
+    rclpy.init()
+
+    node = KickBallNode()
+
+    # MultiThreadedExecutor so the kick action's timed execute callback (its own
+    # reentrant callback group) runs concurrently with the 50 Hz control timer.
+    executor = MultiThreadedExecutor()
```
Flova, line 117: This is really dangerous. For both performance and reliability reasons never use the multi threaded executor. You can use async for the action server callback and have a future 
 that is awaited inside it. This frees the executor while it is waiting for the kick to finish. Then you have a tick / update that is called by the main policy timer. This checks a shared reference of the action goal for things like cancelation etc, sets feedback and sets the result of the future if we are finished/canceled and want to continue with the action callback to the end. This way you gain the advantage of a much more controlled update schedule with significant lowers risks for deadlocks and improved performance.

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
```diff
@@ -6,7 +6,7 @@
       <!-- damping = 10Nm / (155rpm - 10 rpm (in rad/s)) -->
     </default>
     <default class="pi_joint_leg">
-      <joint damping="1.5" actuatorfrcrange="-20 20" armature="0.01316"/>
+      <joint damping="0.8" actuatorfrcrange="-20 20" armature="0.01316"/>
```
Flova: Does the rest of the software still work with this? My policies for example are trained with a different damping.

Flova (changes requested):

src/bitbots_motion/bitbots_hcm/src/hcm.cpp
jgueldenstein: The state is changed to kicking after the a tick of the hcm dsd

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
jgueldenstein: no, it probably is base link projected to ground plane in training, #956

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
jgueldenstein: i changed to always use bitbots tf buffer, as it should be availably, behavior does it like that too

src/bitbots_motion/bitbots_rl_motion/handlers/ball_handler.py
jgueldenstein: the ball handler is actually not used, just the soccer command handler, I fixed it and will retest soon

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: maybe future work? I think this is a bit out of scope for the competition, the way it is done right now is working quite well

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: I have no idea what that is

src/bitbots_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: ball handler removed, changed this to use filtered ball

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: fixed, ball is not available or can not be tfed -> abort kick

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: serious issue during robocup?

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: i fixed it so it should break if kick odom can not be used, a bit risky, if tfs are not availible because motors lag, but should be fine for the moment

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/handlers/soccer_command_handler.py
jgueldenstein: please make a future work issue

src/bitbots_motion/bitbots_rl_motion/bitbots_rl_motion/nodes/kick_ball_node.py
jgueldenstein: please make an issue for future work, the kick works reasonably well with this setup

src/bitbots_simulation/bitbots_mujoco_sim/xml/pi_plus.xml
jgueldenstein: reverted

src/bitbots_misc/bitbots_bringup/launch/simulator_teamplayer.launch
jgueldenstein: resolved in 12e9911ff

PR #938, opened 2026-06-29
Team Comm Limits

### LINKED ISSUE #748: Adjust Team Communication to new rules

cleWu03:
# Summary
We need to make sure we don't break with some team com limit rules

## Proposed changes
Hard limits for message number and message size

## Related issues
replaces pull request #780

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova (approved):

MegaIng (changes requested):
Needs to be adjusted for game_controller_hsl changes

src/lib/game_controller_hsl/game_controller_hsl/scripts/sim_gamestate.py
```diff
@@ -65,6 +65,8 @@ class SimGamestate(Node):
 Stopped:
 
 Goals(Own : Rival):
+Seconds remaining:
```
texhnolyze: With the additional lines I also added the `sim_gamestate.py` script already fill a full screen. Not sure how to handle this best.

MegaIng (approved):

PR #949, opened 2026-06-30
Get Simulation to work with multiple robots

MegaIng:
# Summary

Contains various fixes to make the simulation run with multiple robots, behavior & teamcomm. Works on my laptop

## Proposed changes

- Make sure that each robot reads a different game_settings config for the parameter blackboard so that they don't share a bot_id (and therefore IP address/ports) and role. Indexed by the `ROS_DOMAIN_ID`. Currently 4 configs are there - more could be added if desired. This required a move from an xml to a python launchfile.
- Fix camera topics
- Make the `auto` setting for team com ip automatically select the loopback IP if we are running in simulation
- Work around an issue with the builtin timers in simulation by manually implementing timer logic

## Related issues

Replaces #924 

## Checklist

- [x] Run `pixi run build`
- [ ] ~~Write documentation~~
- [x] Test on your machine
- [ ] ~~Test on the robot~~


src/bitbots_misc/bitbots_parameter_blackboard/config/sim_game_settings_11.yaml
```diff
@@ -0,0 +1,8 @@
+parameter_blackboard:
+  ros__parameters:
+    bot_id: 1
```
Flova, line 3: You can join configs. I would prefer a base config and robot specific overrides so nobody forgets to add a new param to all these configs.

Flova (approved):

src/bitbots_misc/bitbots_parameter_blackboard/config/sim_game_settings_11.yaml
MegaIng: Done. I tried something similar yesterday (using just a dictionary of parameters to override the file) and that didn't work, so I thought it wont work with multiple files either.

src/bitbots_misc/bitbots_parameter_blackboard/launch/parameter_blackboard.launch.py
```diff
@@ -0,0 +1,54 @@
...
+        {"simulation_active": sim},
+        {"use_sim_time": sim},
+        {"field.name": field_name},
+        ParameterFile(PathJoinSubstitution([package_share, "config", "fields", field_name, "config.yaml"])),
+        ParameterFile(PathJoinSubstitution([package_share, "config", "global_parameters.yaml"])),
+        ParameterFile(PathJoinSubstitution([package_share, "config", "game_settings.yaml"])),
+    ]
+
+    robot_domain = os.environ.get("ROS_DOMAIN_ID")
+    if robot_domain is not None:
```
jaagut: Shouldn't we check here whether we use sim?

Flova: Yep. The robot uses a domain id too.

Flova:
The config LGTM. But I am not sure how exactly the python launch handles the overrides, so feel free to give it a quick test.

Flova (approved):

jaagut (approved):

PR #950, opened 2026-06-30
Deploy tool pi plus robots migration to `bitbots` user

texhnolyze:
# Summary

Set our deploy script to use the `bitbots` user and setup in `~/bitbots_main` as `~` is already `/home/bitbots` thenn.

## Related issues

This should be done after https://github.com/bit-bots/ansible/pull/62 is done and deployed to all robots.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot

### DIFF
--- a/scripts/deploy/known_targets.yaml
+++ b/scripts/deploy/known_targets.yaml
@@ -1,22 +1,3 @@
-# "172.20.1.11":
-#   hostname: "nuc1"
-#   robot_name: "amy"
-# "172.20.1.12":
-#   hostname: "nuc2"
-#   robot_name: "rory"
-# "172.20.1.13":
-#   hostname: "nuc3"
-#   robot_name: "jack"
-# "172.20.1.14":
-#   hostname: "nuc4"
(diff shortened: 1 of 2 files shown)


scripts/deploy/known_targets.yaml
```diff
@@ -31,30 +12,50 @@
...
+  # robot_name: "Kalliope"
+# "10.0.6.2":
+  # hostname: "mickey"
+  # robot_name: "Mickey"
+# "10.0.6.3":
+  # hostname: "pink"
+  # robot_name: "Pink"
+# "10.0.6.4":
+  # hostname: "romeo"
+  # robot_name: "Romeo"
```
jaagut: Why did you comment out these?

texhnolyze: Because they will not be up to date anymore, but might be nice to have as a reference at a later stage if we want to revert to the old IP adresses.

Flova (approved):

PR #954, opened 2026-07-01
Deploy: Implement internet connectivity checks

jaagut:
# Summary

- Checks whether the robot can reach data.bit-bots.de (`curl -sSfLI --max-time 5 https://data.bit-bots.de/conda-misc/output/`)
  - If connection can be established, `pixi install` as normal
  - If no connection, use `pixi` with `--as-is` to prevent install failures

Why? We now have a default gateway configured on the robots.
Thus, pixi thinks, we should have an internet connection and tries to sync/install on every command and fails, if there is no true internet connection.

## Checklist

- [X] Run `pixi run build`
- [ ] Write documentation
- [X] Test on your machine
- [X] Test on the robot


Flova:
Is there no other way around this? This seems very risky, as it is not refusing do build without an up to date environment. I am totally for being able to override this behavior, but I have the concern that this leads to outdated/diverging robot env and weird behavior really quickly.

I also don't understand why it wants internet access for an up to date lock file?

Flova:
Maybe the frozen flag is more appropriate. Imo we should always use --frozen on the robot when using the deployment. If the lock file is out of date we need to regenerate it on the host.

jaagut:
```shell
Update Options:
      --no-install
          Don't modify the environment, only modify the lock file
          
          [env: PIXI_NO_INSTALL=]

      --frozen
          Install the environment as defined in the lock file, doesn't update lock file if it isn't up-to-date with the manifest file
          
          [env: PIXI_FROZEN=]

      --locked
          Check if lock file is up-to-date before installing the environment, aborts when lock file isn't up-to-date with the manifest file
          
          [env: PIXI_LOCKED=]

      --as-is
          Shorthand for the combination of --no-install and --frozen
```
We assume, pixi errors "correctly" when the lock doesn't match the env and requires updates

jaagut:
@Flova 
> I am totally for being able to override this behavior
What do you mean by that?

Flova:
That you should be able to e.g. manually override using as-is if you are sure that nothing significantly changed.

Flova:
Maybe we also need locked.

jaagut:
> That you should be able to e.g. manually override using as-is if you are sure that nothing significantly changed.

You want an argument to do the normal install behavior without `--as-is` independent of the current internet connection?

Flova:
> > That you should be able to e.g. manually override using as-is if you are sure that nothing significantly changed.
> 
> You want an argument to do the normal install behavior without `--as-is` independent of the current internet connection?

No. I mean I think as-is is okay to have as long as it is not the default and used with caution.

Flova:
After talking with Jan we came to the conclusion that we probably want to have --frozen and --locked always on the robot by default. --no-install is the one that's dangerous imo and --as-is combines --frozen and --no-install. Why pixie tries to regenerate the lock file is a different story and I suspect it may be related to the way bitbots_rust_nav is compiled. Sometimes the hash of the build changes for whatever reason triggering a rebuilt and update of the lock file. There is a way around this by building it with colcon, which is cleaner imo. that is pr #X with bitbots_cargo, which works on my machine, but had some llvm compatibility error on Cornelius one.

MegaIng:
@jaagut AFAICT from the above discussion, this PR requires further work and may even depend on #944  so I am moving it out of "In Review".

PR #955, opened 2026-07-01
Feature/game controller hsl cleanup

texhnolyze:
# Summary

Make architecture cleaner, by not having different components rely on `game_controller_hsl`.

## Proposed changes

* fix(scripts): automatically select pixi env

based on `IS_ROBOT` environment variable

* refactor(game_controller_hsl): extract response to own component

after properly setting up a PlayerStateResponse in
https://github.com/bit-bots/game_controller_hsl/pull/1

This makes the architecture much cleaner as we do not have dependencies
to `game_controller_hsl` in multiple components, which could also
function without any game controller.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Te st on the robot


src/bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
```diff
@@ -146,8 +140,6 @@ def ball_callback(self, msg: BallArray) -> None:
                 # Store the ball measurement
                 self.ball_state_position = numpify(ball_measurement_map.point)
                 self.ball_state_covariance = covariance
-                # @TODO: actually give last ball time
```
Flova, line 149: We don't publish this anymore right?

src/bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
```diff
@@ -55,14 +55,14 @@ bitbots_ball_filter:
         description: 'Ball measurement certainty in filter'
         validation:
           bounds<>: [0.0, 1.0]
-      
-      distance_factor: 
+
+      distance_factor:
```
Flova, line 59: Are these whitespace changes?

Flova (dismissed):
This changes a lot of things and we need to also adapt all downstream subscriptions of e.g. game state. Otherwise the behavior and localization will crash.

MegaIng:
This PR works, in simulation, as far as I can test it. (which means only  a single robot because the game controller software doesn't work otherwise)

Flova:
Then I am missing something. Because e.g. the penalty field has a new type and has no downstream adaptions. Same with secs remaining. I am pretty sure all of these are referenced in localization, behavior.

src/bitbots_world_model/bitbots_ball_filter/config/ball_filter_parameters.yaml
texhnolyze: Yeah these are only changes of deleting empty whitespace, which my neovim does automatically. I've changed nothing else.

texhnolyze:
The changes to the GameState are as follows:
- The only place where `secs_remaining` were ever used was in `src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/game_status_capsule.py`, which I adapted.
  Maybe this was used in the past in other components, but is not currently.
- The `warnings` field was removed, but was also never used in our code
- The `seconds_till_unpenalized` field changed from `uint16` to `uint8` in accordance with the game controller protocol, but I did not find any place where it was actually used
- The `message_budget` field changed from `uint8` to `uint16` in accordance with the game_controller protocol, but I also did not find any place where it was used yet
- The penalty field was newly added to allow teams to track the penalty state. So far we've only ever checked if we are `penalized`/`penalized_in_place` in behavior or localization, but not for which reason.

src/bitbots_world_model/bitbots_ball_filter/bitbots_ball_filter/ball_filter.py
texhnolyze: Yeah, we only publish it in the aggregator, based on the `/ball_position_relative_filtered` and calculate this time based on the current time and timestamp in the header in `game_controller_hsl`.

jgueldenstein (approved):
tested in sim with multiple robots, and on a single real robot

Flova:
> The changes to the GameState are as follows:
> - The only place where `secs_remaining` were ever used was in `src/bitbots_behavior/bitbots_blackboard/bitbots_blackboard/capsules/game_status_capsule.py`, which I adapted.
>   Maybe this was used in the past in other components, but is not currently.
> - The `warnings` field was removed, but was also never used in our code
> - The `seconds_till_unpenalized` field changed from `uint16` to `uint8` in accordance with the game controller protocol, but I did not find any place where it was actually used
> - The `message_budget` field changed from `uint8` to `uint16` in accordance with the game_controller protocol, but I also did not find any place where it was used yet
> - The penalty field was newly added to allow teams to track the penalty state. So far we've only ever checked if we are `penalized`/`penalized_in_place` in behavior or localization, but not for which reason.

My bad. I missread the diff and thought penalized was removed. Sorry

PR #976, opened 2026-07-02
Port vision to RF DETR

Flova:
# Summary

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot


Flova:
Still needs proper model deployment before merge

Flova:
The latest model looked pretty good. It is also now distributed via conda, so this should be deployment ready.

texhnolyze (approved):
We tested on the robot and it seems to work fine 👍

PR #981, opened 2026-07-03
Fix/quick change

jaagut:
# Summary

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/src/bitbots_misc/bitbots_extrinsic_calibration/config/mickey.yaml
+++ b/src/bitbots_misc/bitbots_extrinsic_calibration/config/mickey.yaml
@@ -1,11 +1,11 @@
 /bitbots_extrinsic_camera_calibration:
   ros__parameters:
-    offset_x: 0.02
+    offset_x: 0.43
     offset_y: 0.0
     offset_z: 0.0
 
 /bitbots_extrinsic_imu_calibration:
   ros__parameters:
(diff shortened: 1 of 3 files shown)


src/bitbots_misc/bitbots_extrinsic_calibration/config/mickey.yaml
```diff
@@ -1,11 +1,11 @@
 /bitbots_extrinsic_camera_calibration:
   ros__parameters:
-    offset_x: 0.02
+    offset_x: 0.43
```
Flova, line 3: Is this still needed?

Flova:
I think the supporter is now positioned behind the ball, so we need to be careful that it doesn't get tangled in the goal.

src/bitbots_misc/bitbots_extrinsic_calibration/config/mickey.yaml
MegaIng: Yes. The head has not been corrected. We will merge this for now, we can undo this once it's changed.

MegaIng:
We will add a min x for the supporter position in a separate PR to prevent it from going into the net.

MegaIng (approved):

PR #982, opened 2026-07-03
Add supporter_min_x and make both parameters relative

MegaIng:
# Summary

Since the supporter position may now be used very close to the goal, we need to keep it away from the goal line such that it doesn't get caught in the net or similar.

## Proposed changes

Add a supporter_min_x parameter. At the same time we also noticed that the max_x was an absolute value despite it probably being supposed to match the field length, meaning depending on the field we may have issues. @Flova Input on this would be nice.

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
Why has this been closed?

MegaIng:
- It would have been a quick fix for a potential issue that we didn't end up using after discussion with Flo
- The better formation system is going to obsolete this

PR #986, opened 2026-07-03
Add increased rate for team comm when in kicking mode

MegaIng:
# Summary

Parts of the behavior rely on knowing if other robots are currently missing. This information getting delayed or getting missed completly because it was less than ~0.5 seconds could be quite bad for the other teammates.

## Proposed changes

The idea is to increase the rate at which we publish when in action kicking. The current parameters increase the check-rate twenty fold, but only actually send when changing to kicking mode. Publishing all the time during kicking would also be an option which a parameter can enabled, but this risks eating a signficant chunk of our message budget. (with current settings, we have 180 seconds left from the message budget for kicking at max rate publishing)

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [ ] Test on the robot


jaagut (changes requested):
Timer should call `high_rate_cb`

jaagut (approved):
@MegaIng Note: always_publish_during_kick is False. Is that intended?

PR #987, opened 2026-07-03
fix(head): offset position on new robots received by hightorque

texhnolyze:
which have an offset in the 0° position in comparison to our previous
robots. As we have never recalibrated any of the Pi+ motors before
and also currently do not have the motor IDs to be able to only
reset the head pitch joint we just add a configurable offset on
`ROBOT_NAME` basis.

If possible the underlying issue should definitely be rectified instead
of relying on this.

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
All our robots now from the newer revision and don't need the head offset angle plate anymore. We have already reset the URDF and mujoco model. But we still need to reset the head-mover bounds.

PR #998, opened 2026-07-04
feat(whistle): changed frequency range to be closer to what the RC26 …

ChlukasX:
# Summary
Whistle detection is unreliable in the RoboCup environment with a high noise floor. Further the whistle being used during games has a set electric noise so we can check the frequency range tighter. The whistle has a concentrated frequency of ~3428Hz, measured with the phyphox app.

This change does not fix the false positives that appear from whistles on other fields.


jaagut:
Did you test this on the robots? Is lowering the threshold necessary?

jaagut (approved):

PR #1006, opened 2026-07-04
Make sure nodes are using sim time if running in sim

MegaIng:
# Summary

Various nodes, especially the ones used internally by the custom transform buffers were not using sim time when running in sim. This almost surely has lead to various issues with additional transform errors that would not have happened IRL.

## Proposed changes

I went through all nodes and made sure they use `use_sim_time=true` if in the correct context.

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,8 @@ commands may use incompatible installations or incomplete environment state.
 - Reuse existing launch patterns and substitutions. Keep launch argument,
   parameter, topic, and namespace names consistent across launch files, config
   files, and node declarations.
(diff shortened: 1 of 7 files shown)


Flova (approved):

Flova:
We could also add it to the AGENTD.md or some other place so LLM reviews keep an eye on that in the future.

jaagut:
@Flova please have a look at the latest commits

Flova:
I don't think the explicit Claude MD is needed. Claude loads the agents MD just fine.

PR #1007, opened 2026-07-04
decide for kicking  depending on distance of seen robots

cleWu03:
# Summary

dribble if there are other robots close, since kicking tools to long

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot
- [ ] Create issues for future work
- [ ] Triage this PR and label it

PR #1013, opened 2026-07-07
Final WM software

MegaIng:
# Summary

This branch represents the software stand as used at the start of the final (mostly). This shouldn't be merged directly, instead it's component PRs should be reviewd & merged.

Further changes that were during the final:

- Before the match, zenoh config was changed so that hopefully no broadcasts are send - this was probably unnecessary, but it's probably valuable to recheck that the robots are not sending anything. From me running wireshark before that change was made, I could not see any other messages send by our robots.
- Before the match we reduced the factor introduced in #986 to 1. This may have increased the actual rate at which teamcom was send/received, although there may have been external factors - it's worth investigating what is happening here.
- During the half time, we disabled #1008 again. It probably makes sense to just not merge that PR.

Component PRs:

- #925 
- #1007 
- #1008 
- #1010


MegaIng:
Branch got renamed to `wm2026_final`

PR #1015, opened 2026-07-08
Migrate ROS2 launch files to better_launch

Flova:
# Summary

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
Before merging, please resolve the better_launch version and packaging discrepancies:

- The vendored upstream revision is `99aea1704`, four commits after release `v1.6.0`, although `package.xml` and `setup.py` still report `1.0.2`. Current upstream `main` (`6954f0ca`) is another four commits ahead.
- Our copy has local patches in `node.py` and `launcher.py`, including `pass_launch_func_default=False`, which two launch files depend on. These patches are not in current upstream, so stock upstream is not a drop-in replacement.
- I found no documented rationale for vendoring. No matching package was found in this workspace’s configured Conda channels, but the Bit-Bots custom Conda recipe already targets a different revision (`bbb904c9`) labelled `1.0.4`.

Please agree on a canonical revision, reconcile the version metadata, document or upstream the required patches, and decide whether to retain the subrepo or provide a patched Conda package through Pixi for both architectures.

_Posted on Jan’s behalf by Codex (GPT-6), based on repository and upstream inspection._

MegaIng:
The patches were done to fix annoyances with the console output of better launch, as well as reduce the chance of introducing accidental global dependencies:

- Better launch defaults to providing all arguments passed to the main launch file to all children. While this is useful for e.g. `sim`, it risks introducing surprising changes in behavior via hidden connections, so we added a default option to disable it.
- By default, better launch logs all commands executed at log level `info`. This makes the output unusable because we e.g. transfer the robot model file via parameters. This was worked around by limiting the length of output, but tbh this is not an optimal solution
- By default, it also `info`s the output of all `exec` commands. This again makes the log output unusable because of the robot model which gets transformed via `xacro`. This was solved by moving the logging here into `debug`
- Better launch was using the python standard log level name `WARNING`, which doesn't work for the ros logger.


I have not checked what kind of changes were made upstream, maybe some or all of these issues were already resolved. I don't think vendoring is absurd since it seems likely that we will need to make further patches to keep the software usable for us, but upstreaming these patches also isn't a bad idea.

PR #1021, opened 2026-07-17
build: use Ninja for workspace builds

jaagut:
# Summary

Switch the default and robot CMake generator from Unix Makefiles to Ninja.

## Proposed changes

- Use the already locked Ninja dependency for default workspace builds.
- Use the same generator for robot workspace builds.

## Impact

Ninja reduces build orchestration overhead and can improve parallel scheduling without changing the compiler or generated runtime code. Existing Makefile build directories must be cleaned once before using this branch.

## Stack

This PR is based on `agent/build-benchmark-baseline` so its CI timings can be compared with the baseline PR.

## Validation

- Parsed `pixi.toml` with Python's TOML parser.
- Listed Pixi tasks successfully.
- Built `bitbots_head_mover` successfully with the Ninja generator.
- `git diff --check`


Flova:
We intentionally opted out of ninja as there where some bugs and due to a controversial opinion of its creator all logs are written to stderr which makes the build really verbose if we want to see build errors at all.

PR #1023, opened 2026-07-17
build: cache compiler outputs with ccache

jaagut:
# Summary

Enable ccache for C and C++ workspace builds and persist compiler outputs in the benchmark workflow.

## Proposed changes

- Add ccache to the shared ROS Pixi feature so it is present in both default and robot environments.
- Configure CMake's C and C++ compiler launchers to use ccache.
- Cache ccache data independently per runner OS, architecture, and lockfile.
- Print compiler cache statistics in the benchmark workflow.
- Regenerate `pixi.lock` for both supported architectures and environments.

## Impact

Repeated builds can reuse compiler output without changing the generated program. The first cold build remains representative, while subsequent CI runs and the touched-source incremental build can report cache hits.

## Stack

This PR is based on `agent/build-benchmark-ninja`.

## Validation

- Located ccache in the configured Conda channels.
- Installed and ran the locked ccache version in the default Pixi environment.
- Built `bitbots_head_mover` successfully through the Ninja and ccache launchers.
- Confirmed ccache recorded compiler calls.
- Parsed `pixi.toml` and ran `git diff --check`.

### DIFF
--- a/.github/workflows/build-benchmark.yml
+++ b/.github/workflows/build-benchmark.yml
@@ -12,6 +12,7 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm]
     runs-on: ${{ matrix.os }}
     env:
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
       PSModulePath: ""
 
     steps:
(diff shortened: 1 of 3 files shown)


Flova:
While faster build times would be nice I am worried about the hidden state it introduces and that a clean build for example might fail even if the ci on a pull request passed.

PR #1029, opened 2026-08-10
Vendor ros launch to fix zombie node error

ayin21:
Because this bug is not getting any attention: https://github.com/ros2/launch/pull/901


jaagut (approved):

jaagut:
Please create an issue to revert this, once https://github.com/ros2/launch/pull/996 lands in jazzy (hopefully) or we upgrade to another ros major version.

PR #1055, opened 2026-09-05
Fix: Not only config striker searches the ball

cleWu03:
# Summary

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


MegaIng (approved):
`CountActivePlayersWithoutGoalie` is long term probably the incorrect tool here, just conceptually. `strategy.role` should be the dynamic role (it isn't yet) and at that point the corresponding check would not do the right thing. But this isn't going to be an issue because the decision doesn't count the current robot anyways, so it doesn't see itself (which also means that it's redundant to use the `WithoutGoalie` decision since only the Goalie does this).

I would still merge this like this. We need to do a pass to rework the role handling anyway.

PR #1072, opened 2026-09-07
Add active vision

Flova:
# Summary

Adds a sampling based active vision algorithm 

## Proposed changes

Samples head targets and scores them based on some cost function including ball, field, robot coverage.

## Related issues

Superceded #856

## Checklist

- [x] Run `pixi run build`
- [ ] Write documentation
- [x] Test on your machine
- [x] Test on the robot

We tested this on the IFA 2026


src/bitbots_motion/bitbots_head_mover/config/head_config.yml
```diff
@@ -196,3 +196,239 @@ move_head:
         validation:
           bounds<>: [0.0, 1.0]
 
+
+  # Sampling based head control used by the ACTIVE_VISION head mode
+  active_vision:
+    root_link:
+      type: string
+      default_value: "base_link"
+      description: "Link the sampled camera poses are resolved against. The map to this link transform places the robot on the field."
```
jaagut, line 205: `The map to this link transform places the robot on the field.` This sentence is grammatically wrong or just really hard to parse.

src/bitbots_motion/bitbots_head_mover/config/head_config.yml
```diff
@@ -196,3 +196,239 @@ move_head:
...
+
+    tip_link:
+      type: string
+      default_value: "camera_optical_frame_left_uncalibrated"
```
jaagut: Why do we use uncalibrated here?

src/bitbots_motion/bitbots_head_mover/config/head_config.yml
```diff
@@ -196,3 +196,239 @@ move_head:
...
+        type: double
+        default_value: 0.3
+        description: "Joint space distance (radians) within which the controller ramps the head speed down linearly from the maximum to zero, so the head glides to a stop on the target. Beyond it the head travels at full speed."
+        validation:
+          gt<>: [0.0]
+
+      control_period:
+        type: double
+        default_value: 0.05
+        description: "Time step (seconds) the controller advances the setpoint by each cycle. Should match the control loop period."
```
jaagut, line 247: We should make it match

src/bitbots_motion/bitbots_head_mover/config/head_config.yml
```diff
@@ -196,3 +196,239 @@ move_head:
...
+        type: double
+        default_value: 0.1
+        description: "Cost per radian of joint space distance between a candidate target and the previously selected one, subtracted from the score. This is the knob that trades reactivity against steady head motion: a larger value keeps the head from jumping between equally attractive parts of the field."
+        validation:
+          gt_eq<>: [0.0]
+
+    debug:
+      enabled:
+        type: bool
+        default_value: true
```
jaagut, line 426: Should probably get turned off

jaagut (changes requested):
Written by GPT-6 on my behalf.

The sampling approach and separation of the planner from ROS integration make sense. Before merging, I’d like to address the following integration issues.

1. **Make field-dimension retrieval asynchronous.**  
   The retry timer calls `get_parameters_from_other_node()`, which internally uses `spin_until_future_complete()` on the head-mover node. After startup, that node already belongs to the EventsExecutor, so the retry throws when it attempts to attach it to another executor. Starting the parameter blackboard late therefore leaves active vision disabled. The separate timer also runs on the same executor thread, so its synchronous wait still blocks head updates.  
   Please use an asynchronous request with a response callback and bounded retries. Add a regression test where the parameter blackboard becomes available after the initial attempt. [Relevant code](https://github.com/bit-bots/bitbots_main/blob/9f55d0e50698b9b24b4cf612fd704bd33da551ac/src/bitbots_motion/bitbots_head_mover/src/move_head.cpp#L239)

2. **Synchronize parameter access across the two executor threads.**  
   `behave()` replaces `params_` while the explicitly created detection thread reads `params_.active_vision.map_frame`. The listener’s internal mutex does not protect this separate copy. This can race during runtime parameter updates, even when only a scoring weight changes.  
   Please give detection callbacks a synchronized configuration snapshot and release the lock before blocking TF lookups. The default EventsExecutor serializes its own callbacks, but it does not serialize them with the added detection executor. [Relevant code](https://github.com/bit-bots/bitbots_main/blob/9f55d0e50698b9b24b4cf612fd704bd33da551ac/src/bitbots_motion/bitbots_head_mover/src/move_head.cpp#L759)

3. **Give the timer and controller one source for their period.**  
   The timer uses a hardcoded `50ms`, while `controller.control_period` is independently configurable. Changing that parameter changes the setpoint step without changing the update frequency.  
   Please derive the controller period from the duration used to create the timer, removing the separate parameter unless changing both together is supported. [Relevant code](https://github.com/bit-bots/bitbots_main/blob/9f55d0e50698b9b24b4cf612fd704bd33da551ac/src/bitbots_motion/bitbots_head_mover/src/move_head.cpp#L356)

4. **Make the exposed parameters match what is actually applied.**  
   `sampling.random_seed` is never passed to the sampler. Coverage `margin`, `cell_size`, and `half_life` are loaded during field-dimension retrieval but ignored by subsequent parameter refreshes.  
   Please wire the seed through and either implement updates for the coverage settings or declare them read-only. Frame settings should likewise be startup-only unless changing them also rebuilds or invalidates the dependent state. Add focused coverage for the supported update behavior. [Relevant code](https://github.com/bit-bots/bitbots_main/blob/9f55d0e50698b9b24b4cf612fd704bd33da551ac/src/bitbots_motion/bitbots_head_mover/src/move_head.cpp#L344)

5. **Retry search initialization after incomplete joint states.**  
   `prev_head_mode_` is updated before trajectory construction succeeds. If the current joint-state message lacks a head joint, construction fails and later valid messages never trigger another attempt for that mode.  
   Please retry initialization once usable joint data arrives. Test an incomplete message followed by a complete message without another mode change. [Relevant code](https://github.com/bit-bots/bitbots_main/blob/9f55d0e50698b9b24b4cf612fd704bd33da551ac/src/bitbots_motion/bitbots_head_mover/src/move_head.cpp#L1072)

There are also a few optional simplifications:

- Remove `active_vision_hold_position_`; it is written and reset but never read.
- Record coverage before constructing a fully configured scorer, so preparation happens once per cycle.
- Have debug markers use kinematics directly instead of constructing a scorer and accessing the world model.
- Move the unrelated `AGENTS.md` changes out of this PR and update comments that still describe trajectory sampling.

The existing unit tests are useful, but they do not exercise the node-level startup and concurrency paths above.

Issue #1056, opened 2026-09-06
Recheck parameters for better assignment


MegaIng:
# Summary

While #925 was merged, the review comments regarding parameter handling were deferred.

We need to go through and handle them. This includes, but is not limited to

- Adding some more parameters for currently hardcoded values
- Renaming some parameters to have better names
- Providing clearer descriptions for some parameters
- Making parameters relative to the field size where relevant
- Making sure all parameters are exposed in a yaml file
- Potentially: Making sure that all parameters are runtime changeable as ROS parameters
- Potentially: Creating a dynamic viewer that areas defined by the different parameters, independent of the effects on actual robots (showing e.g. the gradients)

PR #1079, opened 2026-09-19
Fix/parameters for positioning capsule

cleWu03:
# Summary

- Rename unclear variables
- Some updates on variable descriptions
- Check for absolute or relative values
- All params included into the body_behavior.yaml

Part of #1056 but doesnt include all the there suggested features. Would be nice to merge it soon anyway to enable working on the positioning capsule without merge conflicts in the future

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


MegaIng (changes requested):
The parameters in body_behavior.yaml should be moved into a block "formation" or "positioning" so that it's clear that they belong together. This then also means "formation_include_supporter".

Also, the list of sliders and the way those values are being read out and put into the params in the debug script should be refactored to be maintainable. It's a large unmaintainable script but by tackling sections as we have to touch them we will get there ;-)

Issue #1057, opened 2026-09-06
We don't search for the ball, if the configured striker is out of game


cleWu03:
# Summary

Only the config striker has the search ball behaviour.

## Expected behavior

The decision who is the searching robot should be dynamic, depending on how close to the opponents goal the robots are.

cleWu03:
With the merge of #1085 the same issue will apply in the #DefensiveSetPlay subtree. Must solve it there too.

PR #1085, opened 2026-09-23
Using positioning only if ball seen

### LINKED ISSUE #1084: Using positioning without ball seen
# Summary

In the #DefensiveSetPlay Tree of the main dsd we always use the GoToFormationposition behavior, even if the ball is bad / not seen 

## Expected behavior

Only go to Formation with a good ball, search for the ball otherwise

## Current behavior

We use a bad ball and the positioning script returns a useless position far away from the correct position close to the sideline and looking out of the field (so the ball doesnt get better over time :) )

cleWu03:
# Summary

Quick solve for #1084. It produces a code duplication of the suboptimal code from #1057. I will comment there, that we remember to duplicate our solution there too. 

## Checklist

- [ ] Run `pixi run build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


src/bitbots_behavior/bitbots_body_behavior/bitbots_body_behavior/behavior_dsd/main.dsd
```diff
@@ -102,7 +102,13 @@ $SecondBallTouchAllowed
 @SetNoSecondBallContactVariable + value:true + r:false, @ForgetBallStartPosition + r:false
 
 #DefensiveSetPlay
-@ChangeAction + action:positioning, @ForgetBall, @LookAtFieldFeatures, @GoToFormationPosition + set_play:true + stand:true + enter_position:%placing_enter_position + leave_position:%placing_leave_position + enter_orientation:%placing_enter_orientation + leave_orientation:%placing_leave_orientation
+$BallSeen
+    NO --> $ConfigRole
+        GOALIE --> $CountActiveRobotsWithoutGoalie
+            ZERO --> #SearchBall
+            ELSE --> #RolePositionWithPause
+        ELSE --> #SearchBall
```
Flova: I think only the striker should search the ball all others should go to the role position like we do during normal play. But I would change it so all robots do the wait + turn for a few seconds, but then only the config striker goes to search the ball and all others do to the role positions. If we add this to the search ball tree we can reuse it here with a simple ball seen decision.

cleWu03: Since a change after the IFA game, where the config striker was out of field and no one else was searching the ball, this is the same behavior as during normal play. #1057 describes this problem further. We should find a solution where the decision who searches is not only based on the config role, to make sure exactly one robot does the search behavior in any cases. I like the idea of a U-Turn for all at the beginning, but would handle that in the fix for the other Issue, since it is not directly related to the problem of using the positioning capsule in a wrong moment

